PR #1, opened 2022-02-24
Condition opcode alternative

EnderShadow:
This is another way of specifying the conditional opcodes. This also assumes that subtract with carry


EnderShadow:
Closed after discussion on discord

PR #6, opened 2022-05-09
Expanded opcodes

EnderShadow:
A few things to think about for this PR.

1. Do we want to remove the expanded registers extension and integrate it through the current reserved bits?
   -  We could also reclaim the prefixes reserved in the `110x xxxx` area and add an additional C or reserved bit if we think that's a good idea.
2. Are `rsbb` and `rcmp` useful?
3. Do we want the shifts and added bitwise functions here or in another extension?

Depending on what you all think, I may or may not move some of the sections around a bit

### DIFF
--- a/extensions/expanded-opcodes/README.md
+++ b/extensions/expanded-opcodes/README.md
@@ -11,77 +11,52 @@ This extension adds an expanded instruction format to allow for a larger number 
 
 # Added Instructions
 
-The following opcodes are now defined.
+## Calculation Instructions
 
-| First byte    | Second Byte  | Third Byte   | Comment                                                                   |
-|:--------------|:-------------|:-------------|:--------------------------------------------------------------------------|
-| `111 00 000`  | `SS CCCCCC`  | `RRR RRR MM` | Register-Register operation                                               |
-| `111 00 ???`  | `SS CCCCCC`  | `RRR RRR MM` | When `???` is not `000`, reserved                                         |
-| `111 01 000`  | `SS CCCCCC`  | `RRR IIIII`  | Register-Immediate operation                                              |
-| `111 01 ???`  | `SS CCCCCC`  | `RRR IIIII`  | When `???` is not `000`, reserved                                         |
-| `111 10 DDD`  | `DDDD DDDD`  | `DDDD DDDD`  | Large relative jump                                                       |
-| `111 11 DDD`  | `DDDD DDDD`  | `DDDD DDDD`  | Large relative call if stack extension is implemented, otherwise reserved |
+| First byte    | Second Byte   | Third Byte   | Comment                      |
+|:--------------|:--------------|:-------------|:-----------------------------|
+| `111 0 CCCC`  | `C 0 SS CCCC` | `RRR RRR MM` | Register-Register operation  |
+| `111 0 CCCC`  | `C 1 SS CCCC` | `RRR IIIII`  | Register-Immediate operation |
 
 | Symbol | Meaning                                    |
 |--------|--------------------------------------------|
-| C      | condition code bit                         |
+| C      | opcode bit                                 |
 | R      | register id                                |
 | I      | immediate                                  |
-| D      | signed displacement                        |
 | S      | size                                       |
 | M      | memory mode                                |
-| ?      | reserved for extensions or another purpose |
 
-## Mapping from the base ISA opcodes to this extension
+### Opcode table
 
-When the base opcode `CCCC` matches `0xxx`, the new opcode is `0 CCCC 0`
-
-When the base opcode `CCCC` matches `1xxx`, the new opcode is `01 CCCC`
-
-## Opcode mapping table
-
-| `CCCCCC` | NAME            | Operation                          | Flags  | Comment     |
-|----------|-----------------|------------------------------------|--------|-------------|
-| `000000` | `ADD`           | `A ← A + B`                        | `ZNCV` |             |
-| `000001` | `ADC`           | `A ← A + B + C`                    | `ZNCV` |             |
-| `000010` | `SUB`           | `A ← A - B`                        | `ZNCV` |             |
-| `000011` | `SBB`           | `A ← A - B - C`                    | `ZNCV` |             |
-| `000100` | `RSUB`          | `A ← B - A`                        | `ZNCV` | (1)         |
-| `000101` | `RSBB`          | `A ← B - A - C`                    | `ZNCV` |             |
-| `000110` | `CMP`           | `_ ← A - B`                        | `ZNCV` |             |
-| `000111` | `RCMP`          | `_ ← B - A`                        | `ZNCV` |             |
-| `001000` | `OR`            | <code>A ← A &#124; B</code>        | `ZN`   | (5)         |
-| `001001` | `NOR`           | <code>A ← ~(A &#124; B)</code>     | `ZN`   | (5)         |
-| `001010` | `XOR`           | `A ← A ^ B`                        | `ZN`   | (5)         |
-| `001011` | `XNOR`          | `A ← ~(A ^ B)`                     | `ZN`   | (5)         |
-| `001100` | `AND`           | `A ← A & B`                        | `ZN`   | (5)         |
-| `001101` | `NAND`          | `A ← ~(A & B)`                     | `ZN`   | (5)         |
-| `001110` | `TEST`          | `_ ← A & B`                        | `ZN`   | (5)         |
-| `001111` |                 |                                    |        | reserved    |
-| `010000` | `SHL`           | `A ← A << B`                       | `ZN`   | (5)         |
-| `010001` | `ROL`           | `A ← A rotate left B`              | `ZN`   | (5)         |
-| `010010` | `LSHR`          | `A ← A >>> B`                      | `ZN`   | (5)         |
-| `010011` | `ASHR`          | `A ← A >> B`                       | `ZN`   | (5)         |
-| `010100` |                 |                                    |        | reserved    |
-| `010101` |                 |                                    |        | reserved    |
-| `010110` |                 |                                    |        | reserved    |
-| `010111` |                 |                                    |        | reserved    |
-| `011000` | `MOVZ`          | `A ← B`                            | None   | (7)         |
-| `011001` | `MOV` or `MOVS` | `A ← B`                            | None   | (8)         |
-| `011010` | `LOAD`          | `A ← MEM[B]`                       | None   |             |
-| `011011` | `STORE`         | `MEM[B] ← A`                       | None   | (2)         |
-| `011100` | `SLO`           | <code>A ← (A << 5) &#124; B</code> | None   | (3) (6)     |
-| `011101` |                 |                                    |        | reserved    |
-| `011110` | `READCR`        | `A ← CR[B]`                        | None   | (4) (6)     |
-| `011111` | `WRITECR`       | `CR[B] ← A`                        | None   | (2) (4) (6) |
-| `1xxxxx` |                 |                                    |        | reserved    |
+| `C CCCC CCCC` | NAME            | Operation                          | Flags  | Comment     |
+|---------------|-----------------|------------------------------------|--------|-------------|
+| `0 0000 0000` | `ADC`           | `A ← A + B + C`                    | `ZNCV` | (2)         |
+| `0 0000 0001` | `SBB`           | `A ← A - B - C`                    | `ZNCV` | (2)         |
+| `0 0000 0010` | `RSBB`          | `A ← B - A - C`                    | `ZNCV` | (1) (2)     |
+| `0 0000 0011` |                 |                                    |        | reserved    |
+| `0 0000 01xx` |                 |                                    |        | reserved    |
+| `0 0000 1xxx` |                 |                                    |        | reserved    |
+| `0 0001 xxxx` |                 |                                    |        | reserved    |
+| `0 001x xxxx` |                 |                                    |        | reserved    |
+| `0 01xx xxxx` |                 |                                    |        | reserved    |
+| `0 1xxx xxxx` |                 |                                    |        | reserved    |
+| `1 xxxx xxxx` |                 |                                    |        | reserved    |
 
 
-1) Enables NEG and NOT to be encoded as `RSUB r, imm`.
-3) Designed to allow for building a larger immediate value. To reach a full 16 bit immediate, a 4th `SLO` or an additional `NOT` may be required.
-4) Primary intent is that these are used with immediate. Exact assignment of control registers is still floating. At least the the CPU status/extension/feature control registers should be present.
-5) The C and O flags are in an undefined state after execution of these instructions. Implementations may do whatever is easiest. An extension may mandate a particular behavior, with good enough reason, but must *not* mandate that the value of these flags after the operation depends on their value before the operation.
-6) These instructions do not have a 2 register mode.
-7) This instruction zero extends argument B as if the value read is of the size specified by the SS bits in the instruction (assuming the relevant feature is present)
-8) This instruction sign extends argument B as if the value read is of the size specified by the SS bits in the instruction (assuming the relevant feature is present)
+1) Enables NEG and NOT to be encoded as `RSBB r, imm` for integers larger than supported by the ISA.
+2) C in the operation refers to the carry bit.
+
+## Jump and Call
+
+| First byte   | 2nd-9th Byte | Comment                                                             |
+|:-------------|:-------------|:--------------------------------------------------------------------|
+| `111 10 JJJ` | `DDDD DDDD`  | Relative jump                                                       |
+| `111 11 JJJ` | `DDDD DDDD`  | Relative call if stack extension is implemented, otherwise reserved |
+
+| Symbol | Meaning                                         |
+|--------|-------------------------------------------------|
+| J      | the number of bytes to use for the displacement |
+| D      | signed displacement                             |
+
+`JJJ` is 1 less than the number of bytes to use for the displacement. This means that `JJJ = 000` is 1 additional byte and `JJJ = 111` is 8 additional bytes.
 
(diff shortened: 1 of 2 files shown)


JKTKops:
`rsbb` is theoretically useful for the same reasons as `rsub` (higher precision arithmetic negation). `rcmp` is useless (anything that you could do with `rcmp` can provably be done with just `cmp`).

What things go here depend on what we want this extension to be. Is this "true" ETCa, or is this merely a "common extension"? If it's merely a common extension, I think this extension should probably only spec the format, `adc`, `sbb`, possibly `rsbb` (if we go with that), and the rest should be delegated to small BMEs or other added opcode extensions. If this is "true" ETCa, then this extension should contain all the core operations that are missing from base and nothing more (worth noting that in my view, "true" ETCa does already have VWI, as I consider memory operands part of the true ISA).

I generally favor having lots of small extensions rather than having medium or large extensions that group together things which are conceptually unrelated.

extensions/expanded-opcodes/README.md
```diff
@@ -0,0 +1,87 @@
...
+# Added Instructions
+
+The following opcodes are now defined.
+
+| First byte    | Second Byte  | Third Byte   | Comment                                                                   |
+|:--------------|:-------------|:-------------|:--------------------------------------------------------------------------|
+| `111 00 000`  | `SS CCCCCC`  | `RRR RRR MM` | Register-Register operation                                               |
+| `111 00 ???`  | `SS CCCCCC`  | `RRR RRR MM` | When `???` is not `000`, reserved                                         |
+| `111 01 000`  | `SS CCCCCC`  | `RRR IIIII`  | Register-Immediate operation                                              |
+| `111 01 ???`  | `SS CCCCCC`  | `RRR IIIII`  | When `???` is not `000`, reserved                                         |
```
JKTKops: For consistency, (ease of decoding, etc.) why not use `111 CCCCC FF SS CCCC`? this makes the location of the FF and SS bits match within the byte, which should reduce the number of multiplexers needed to build efficient decoders.

This part is obviously very in flux anyway, I expect it to change a few times.

extensions/expanded-opcodes/README.md
```diff
@@ -0,0 +1,87 @@
...
+| `111 10 DDD`  | `DDDD DDDD`  | `DDDD DDDD`  | Large relative jump                                                       |
+| `111 11 DDD`  | `DDDD DDDD`  | `DDDD DDDD`  | Large relative call if stack extension is implemented, otherwise reserved |
```
JKTKops: Let's adopt the x86 terms for "short jump/call," "near jump/call," and "far jump/call." These would be "near"; but in x86 the near transfers get a 32-bit displacement which makes me think that this format should be adjusted to be more VWI than this. E.g. `1111 CCCC [4-bytes signed displacement]`. Then CCCC = 1111 could be a call instead, or something. IDK. Some thought should be put into simplifying the decoding across all of the jump representations that we have now.

extensions/expanded-opcodes/README.md
```diff
@@ -0,0 +1,87 @@
...
+| `010110` |                 |                                    |        | reserved    |
+| `010111` |                 |                                    |        | reserved    |
+| `011000` | `MOVZ`          | `A ← B`                            | None   | (7)         |
+| `011001` | `MOV` or `MOVS` | `A ← B`                            | None   | (8)         |
+| `011010` | `LOAD`          | `A ← MEM[B]`                       | None   |             |
+| `011011` | `STORE`         | `MEM[B] ← A`                       | None   | (2)         |
+| `011100` | `SLO`           | <code>A ← (A << 5) &#124; B</code> | None   | (3) (6)     |
+| `011101` |                 |                                    |        | reserved    |
+| `011110` | `READCR`        | `A ← CR[B]`                        | None   | (4) (6)     |
+| `011111` | `WRITECR`       | `CR[B] ← A`                        | None   | (2) (4) (6) |
```
JKTKops: Why does this table have things which can already be encoded with a 1 byte opcode? There is no reason to have a strictly worse encoding for any of them.

extensions/expanded-opcodes/README.md
```diff
@@ -0,0 +1,87 @@
...
+| `CCCCCC` | NAME            | Operation                          | Flags  | Comment     |
+|----------|-----------------|------------------------------------|--------|-------------|
+| `000000` | `ADD`           | `A ← A + B`                        | `ZNCV` |             |
+| `000001` | `ADC`           | `A ← A + B + C`                    | `ZNCV` |             |
+| `000010` | `SUB`           | `A ← A - B`                        | `ZNCV` |             |
+| `000011` | `SBB`           | `A ← A - B - C`                    | `ZNCV` |             |
+| `000100` | `RSUB`          | `A ← B - A`                        | `ZNCV` | (1)         |
+| `000101` | `RSBB`          | `A ← B - A - C`                    | `ZNCV` |             |
+| `000110` | `CMP`           | `_ ← A - B`                        | `ZNCV` |             |
+| `000111` | `RCMP`          | `_ ← B - A`                        | `ZNCV` |             |
```
JKTKops: This should not exist.

extensions/expanded-opcodes/README.md
```diff
@@ -0,0 +1,87 @@
...
+| `001000` | `OR`            | <code>A ← A &#124; B</code>        | `ZN`   | (5)         |
+| `001001` | `NOR`           | <code>A ← ~(A &#124; B)</code>     | `ZN`   | (5)         |
+| `001010` | `XOR`           | `A ← A ^ B`                        | `ZN`   | (5)         |
+| `001011` | `XNOR`          | `A ← ~(A ^ B)`                     | `ZN`   | (5)         |
+| `001100` | `AND`           | `A ← A & B`                        | `ZN`   | (5)         |
+| `001101` | `NAND`          | `A ← ~(A & B)`                     | `ZN`   | (5)         |
```
JKTKops: I'm not convinced that NOR, XNOR, or NAND should have opcodes (and certainly not in this extension, regardless). Other ISAs don't have them.

JKTKops (changes requested):

extensions/expanded-opcodes/README.md
JKTKops: github apparently decided to ignore my line selection on this one: specifically commenting on `rcmp`.

EnderShadow:
> `rsbb` is theoretically useful for the same reasons as `rsub` (higher precision arithmetic negation). `rcmp` is useless (anything that you could do with `rcmp` can provably be done with just `cmp`).

That's what I get for writing this late at night/early in the morning. I'll remove `rcmp`

> What things go here depend on what we want this extension to be. Is this "true" ETCa, or is this merely a "common extension"? If it's merely a common extension, I think this extension should probably only spec the format, `adc`, `sbb`, possibly `rsbb` (if we go with that), and the rest should be delegated to small BMEs or other added opcode extensions. If this is "true" ETCa, then this extension should contain all the core operations that are missing from base and nothing more (worth noting that in my view, "true" ETCa does already have VWI, as I consider memory operands part of the true ISA).
> 
> I generally favor having lots of small extensions rather than having medium or large extensions that group together things which are conceptually unrelated.

I was thinking that this would be more of "true" ETCa, but I'm fine with just making this a common extension

extensions/expanded-opcodes/README.md
EnderShadow: With the shifting of the opcodes so that the base opcodes fit here as well, the least significant 4 bits being `CCCC` doesn't really fit unless you're specifically just moving the FF and SS bits so that they match.

extensions/expanded-opcodes/README.md
EnderShadow: We could always make these 2 instructions use additional bytes to make them far jumps/calls. I don't see a need to move the call to it's own instruction.

extensions/expanded-opcodes/README.md
EnderShadow: That's actually what prompted question number 1 in the PR description. Having them here as a mapping from the base-isa to larger opcodes doesn't make sense if there's no benefit which is why I suggested removing the expanded register extension and merging the larger registers into here. The 3 reserved bits in the PR would go to the registers and may or may not be shuffled around a bit. For the immediate mode, 2 of the bits would go to the immediate instead. Since we'd be removing the expanded registers extension, we'd be able to reclaim that bit as an additional `C` bit (or something else)

extensions/expanded-opcodes/README.md
EnderShadow: Yeah, I'll remove it. I blame doing this at 3 or 4 am :stuck_out_tongue:

extensions/expanded-opcodes/README.md
EnderShadow: It really depends on what bitwise operators we want to have. I just kind of placed them there just because.

extensions/expanded-opcodes/README.md
JKTKops: But there's no real reason for the base opcodes to have to fit here. That's just tossing away 16 opcodes for no advantage (they have to be decoded separately anyway!) And yes, the suggestion is to move those bits so that they match.

extensions/expanded-opcodes/README.md
JKTKops: Well they don't need the 3 displacement bits in the first byte. Perhaps something can be done here to simplify them but idk.

extensions/expanded-opcodes/README.md
JKTKops: The expanded registers being r8-r15? I'd prefer to keep that on the prefix as that bumps us from 64 extra opcodes here, which is definitely not enough to cover the whole future ISA, to 512, which probably is.

extensions/expanded-opcodes/README.md
MegaIng: I agree with Grape here, I prefer `FF SS CCCC` (We will see about `CCCCC` in the first byte when other extensions come I think). I also don't see a reason for the base opcodes to fit here.

extensions/expanded-opcodes/README.md
MegaIng: I also agree with Grape here, keep the extended registers extensions  separate, and don't have the base opcodes here.

extensions/expanded-opcodes/README.md
```diff
@@ -0,0 +1,87 @@
...
+| `011010` | `LOAD`          | `A ← MEM[B]`                       | None   |             |
+| `011011` | `STORE`         | `MEM[B] ← A`                       | None   | (2)         |
+| `011100` | `SLO`           | <code>A ← (A << 5) &#124; B</code> | None   | (3) (6)     |
+| `011101` |                 |                                    |        | reserved    |
+| `011110` | `READCR`        | `A ← CR[B]`                        | None   | (4) (6)     |
+| `011111` | `WRITECR`       | `CR[B] ← A`                        | None   | (2) (4) (6) |
+| `1xxxxx` |                 |                                    |        | reserved    |
+
+
+1) Enables NEG and NOT to be encoded as `RSUB r, imm`.
```
MegaIng: This also applies to `RSBB` if you want to simulate negation of larger than register sizes ints. Not sure if that needs to be mentioned.

extensions/expanded-opcodes/README.md
MegaIng: To note: Having the extended registers prefix has the same byte length as using this extension for the base opcodes.

extensions/expanded-opcodes/README.md
EnderShadow: I couldn't completely preserve the `FF SS CCCC` portion because of the jump and call, but this new version should work.

extensions/expanded-opcodes/README.md
EnderShadow: I think I've come up with a good solution to this. Take a look at the new commit

extensions/expanded-opcodes/README.md
EnderShadow: Alright, I removed the instructions that are in the base ISA from here.

extensions/expanded-opcodes/README.md
EnderShadow: I removed them since the expanded opcodes section is much more lean in terms of added instructions now

extensions/expanded-opcodes/README.md
EnderShadow: I added that as a note. Good catch.

extensions/expanded-opcodes/README.md
```diff
@@ -11,77 +11,52 @@ This extension adds an expanded instruction format to allow for a larger number
...
-| `111 00 000`  | `SS CCCCCC`  | `RRR RRR MM` | Register-Register operation                                               |
-| `111 00 ???`  | `SS CCCCCC`  | `RRR RRR MM` | When `???` is not `000`, reserved                                         |
-| `111 01 000`  | `SS CCCCCC`  | `RRR IIIII`  | Register-Immediate operation                                              |
-| `111 01 ???`  | `SS CCCCCC`  | `RRR IIIII`  | When `???` is not `000`, reserved                                         |
-| `111 10 DDD`  | `DDDD DDDD`  | `DDDD DDDD`  | Large relative jump                                                       |
-| `111 11 DDD`  | `DDDD DDDD`  | `DDDD DDDD`  | Large relative call if stack extension is implemented, otherwise reserved |
+| First byte    | Second Byte   | Third Byte   | Comment                      |
+|:--------------|:--------------|:-------------|:-----------------------------|
+| `111 0 CCCC`  | `C 0 SS CCCC` | `RRR RRR MM` | Register-Register operation  |
+| `111 0 CCCC`  | `C 1 SS CCCC` | `RRR IIIII`  | Register-Immediate operation |
```
JKTKops, line 19: Swap the 0 in the first byte with the C in the second byte. You've split the FF bits across the bytes!

The jumps and calls don't need to be 1 byte opcodes, so it's fine if their FF indicators are in the second byte. See my comments there.

extensions/expanded-opcodes/README.md
```diff
@@ -11,77 +11,52 @@ This extension adds an expanded instruction format to allow for a larger number
...
+
+1) Enables NEG and NOT to be encoded as `RSBB r, imm` for integers larger than supported by the ISA.
+2) C in the operation refers to the carry bit.
+
+## Jump and Call
+
+| First byte   | 2nd-9th Byte | Comment                                                             |
+|:-------------|:-------------|:--------------------------------------------------------------------|
+| `111 10 JJJ` | `DDDD DDDD`  | Relative jump                                                       |
+| `111 11 JJJ` | `DDDD DDDD`  | Relative call if stack extension is implemented, otherwise reserved |
```
JKTKops: I really do not like this `JJJ` thing. It's a heavy decoding cost (not just figuring out how long the instruction is, but also needing to support _handling_ instructions of all those lengths) for very little practical benefit. We could instead have
```
111 ????? 1J ?? CCCC (rmi)|(DDDD)+
```
There's lots of things these `?` bits could be used for to get many different types of jumps and calls; I'd suggest simply forcing them to particular values and letting everything else be reserved. For example, they could make the transfers switch between absolute/relative, they could make the address/displacement come from an `rmi` byte instead of a literal displacement (i.e. from the AAA register or from a memory operand). I wouldn't want to allow displacements of sizes other than 1, 2, and 4 bytes, and absolute addresses should be the same size as the current pointer mode.

With this approach, the conditional prefix should not be allowed to appear on any of these, since they always have a condition code regardless.

JKTKops (changes requested):

extensions/expanded-opcodes/README.md
EnderShadow: I don't really see what the issue with this is. Especially since we're already not using the base isa as a subset of the expanded isa.

extensions/expanded-opcodes/README.md
EnderShadow: If you're really that against being able to use any 1-8 byte displacement, maybe making it a logarithmic thing so that it's 1, 2, 4, and 8 for 2 of the J bits and then the 3rd will select between relative and absolute. I don't see an issue with requiring the conditional prefix to make this conditional since it doesn't make it any shorter.

JKTKops: It's less about not requiring a prefix and more about putting the FF bits in a consistent place. Taking a cue from RISC-V, which goes to some seemingly strange lengths to put common subsets of instructions in the same place.

extensions/expanded-opcodes/README.md
JKTKops: It's about putting the FF bits in a consistent place. Taking a cue from RISC-V, which goes to some seemingly strange lengths to put common subsets of instructions in the same place. The opcode decoder has to be separate, but the same logic can identify the instruction type more easily.

EnderShadow: This still follows that principle. The lone C bit in the second byte is what is normally used for selecting between calculations and jumps/calls.

extensions/expanded-opcodes/README.md
EnderShadow: This still follows that. Just the mode selection is in a different location for expanded opcodes. See my response for the calculation opcodes

extensions/expanded-opcodes/README.md
```diff
@@ -48,15 +48,20 @@ This extension adds an expanded instruction format to allow for a larger number
...
+| `111 11 1 JJ` | `IIII IIII`  | Absolute call if stack extension is implemented, otherwise reserved |
 
 | Symbol | Meaning                                         |
 |--------|-------------------------------------------------|
 | J      | the number of bytes to use for the displacement |
 | D      | signed displacement                             |
+| I      | unsigned immediate                              |
 
-`JJJ` is 1 less than the number of bytes to use for the displacement. This means that `JJJ = 000` is 1 additional byte and `JJJ = 111` is 8 additional bytes.
+`JJ` is log base 2 of the number of bytes to use for the displacement. This means that `JJ = 00` is 1 additional byte and `JJ = 11` is 8 additional bytes.
```
JKTKops: Maybe we should just call these SS bits and say they are the size? "log base 2" is overcomplicated, it would be simpler and more direct to say "`JJ = 00,01,10,11` means there are 1,2,4,8 additional bytes for the displacement or address."

JKTKops (dismissed):

extensions/expanded-opcodes/README.md
EnderShadow: Good idea

MegaIng (approved):

JKTKops (approved):

PR #7, opened 2022-05-09
Add conditional execution prefix to variable width standard as a defined prefix

EnderShadow:
Minor documentation update

### DIFF
--- a/extensions/variable-width-instruction-standard.md
+++ b/extensions/variable-width-instruction-standard.md
@@ -16,4 +16,5 @@ Instruction prefixes are a sequence of one or more bytes which modify the instru
 
 ### Defined Prefix Sequences
 
+`1010 xxxx` when `xxxx` is not `1111` nor `1110`
 `110x xxxx`
--- a/extensions/variable-width-instruction-standard.md
+++ b/extensions/variable-width-instruction-standard.md
@@ -16,5 +16,10 @@ Instruction prefixes are a sequence of one or more bytes which modify the instru
 
 ### Defined Prefix Sequences
 
-`1010 xxxx` when `xxxx` is not `1111` nor `1110`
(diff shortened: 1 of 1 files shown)


extensions/variable-width-instruction-standard.md
```diff
@@ -16,4 +16,5 @@ Instruction prefixes are a sequence of one or more bytes which modify the instru
 
 ### Defined Prefix Sequences
 
+`1010 xxxx` when `xxxx` is not `1111` nor `1110`
 `110x xxxx`
```
JKTKops: Would be nice if this section had a short (<= 1 sentence) summary of what these prefixes are

JKTKops (dismissed):

extensions/variable-width-instruction-standard.md
EnderShadow: Good idea.

extensions/variable-width-instruction-standard.md
```diff
@@ -16,4 +16,10 @@ Instruction prefixes are a sequence of one or more bytes which modify the instru
 
 ### Defined Prefix Sequences
 
-`110x xxxx`
+
+| Prefix      | Description        |
+|:------------|:-------------------|
+| `1010 xxxx` | Unused             |
+| `1100 xxxx` | Expanded registers |
```
MegaIng, line 23: This could link to the corresponding document

MegaIng (approved):

extensions/variable-width-instruction-standard.md
EnderShadow: Possibly. I figured the description would say what uses it. Otherwise it can be put in the VWI standard to keep track of what's a prefix and what isn't as well as what is still unused from the ISA

PR #12, opened 2022-05-17
Advanced operand extensions

JKTKops:
nothing new to say after the hours spent in discord... lots to review though.


extensions/README.md
```diff
@@ -2,7 +2,7 @@
 
 | CPUID bit | Extension                                                 | Dependencies | State             |
 |:---------:|-----------------------------------------------------------|:------------:|-------------------|
-|     0     | Memory Mode Operations + Expanded Immediates              | VWI          | Planned           |
+|     0     | [Full Immediate](./full-immediates)                       | VWI          | Planned           |
```
EnderShadow: This should be changed from Planned to Under Development

extensions/full-immediates/README.md
```diff
@@ -0,0 +1,47 @@
...
+|----|-----------------------|
+| 00 | 1 |
+| 01 | 2 |
+| 10 | 4 |
+| 11 | 4 |
+
+An 8-byte literal immediate cannot be encoded by this extension. However, any encoding
+which "modifies the meaning of `iS`" will affect the `iS` immediate defined here.
+
+TODO: add explicit reference to the `REX.Q` bit once that is specified.
```
EnderShadow: You can make the relevant changes to the register extension in this PR if you want.

extensions/memory-operands-1/README.md
```diff
@@ -0,0 +1,130 @@
...
+  else:
+    M ← RES
+```
+In plain language; when `D` is 0, the `dst` operand is a register and the `src` operand is a memory location.
+When `D` is 1, these are swapped. In all cases, the value in the memory location must be loaded from memory.
+When the `dst` operation is a memory location, the result of the operation must also be stored there.
+
+If an operation uses an `ABM` byte to encode two operands, but only treats one of them as an input, then this
+behavior should likewise apply to memory operations; memory need not be loaded if the memory operand is not an input
+(e.g., as is the case for `mov m, r/i`), however, extraneous memory _reads_ are permitted.
```
EnderShadow: Do we really want to permit this? It could cause issues for MMIO if reads trigger something.

extensions/memory-operands-1/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+The instructions Operand Size (usually from the instruction `SS` bits, but NOT from the `SIB.S` bits) informs the size of the read or writes. These reads
+and writes must behave, with regards to how memory is read and modified, the same as `ld` or `st` instructions
+to the same address, with the same operand size.
+
+When computing an address, register contents must be treated as if they were the width of an
+address in the current address mode. When performing the operation, the operation must be performed
+as though the operand values were the size specified by the instruction's Operand Size attribute.
+
+There is no performance restriction on instructions using these modes. They do not need to operate in
```
EnderShadow: Does this even need to be specified? We've never specified any sort of performance metric for anything before.

extensions/memory-operands-2/README.md
```diff
@@ -37,6 +37,20 @@ Familiarity with the full-immediates extension is encouraged likewise.
 
 The value `ip` is the address in memory of the current instruction.
 
+## `REX.Q`
+
+A `REX.Q` prefix can be used with instructions in this format if all of the following conditions are met:
+  - The instruction has _exactly one_ of a displacement or immediate. It cannot have both.
+  - The instruction is an `iS` mode with Operand Size attribute `quad`
```
EnderShadow: Isn't this a bit contradictory since it says only one of immediate or displacement is allowed but then requires both of them?

extensions/memory-operands-2/README.md
```diff
@@ -0,0 +1,77 @@
...
+  - The instruction is a `dP` mode and the current Addressing Mode is `quad`
+  - The [register expansion](../expanded-registers/README.md) extension is available
+Then the `iS` or `dP` will be an 8-byte value rather than a 4-byte value.
+
+The prefix remains illegal if any of the `REX.A`, `REX.B`, or `REX.X` bits are set and the register expansion extension is not enabled.
+`REX.Q` is **illegal**, not reserved, for instructions in these formats which do not satisfy the above conditions.
```
EnderShadow: Is illegal really the term we want here? Would it be better to say "REX.Q must be 0"?

extensions/memory-operands-1/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+A `dP` is the same size as a logical address in the current mode. However, similarly to [full-immediates](../full-immediates/README.md),
+this extension cannot encode an 8 byte displacement. In `quadword` address mode, a `dP` is 4 bytes.
+
+If the [register expansion](../expanded-registers/README.md) extension is available, then a `REX.Q` prefix may be used with instructions in these formats if
+  - The instruction contains a `dP`
+  - The current Addressing Mode is `quadword`
+In this case, the `dP` is an 8-byte value rather than a 4-byte value.
+
+For other instructions in these formats, `REX.Q` is **illegal**, not reserved.
```
EnderShadow: Similarly, is this the best way to word it?

EnderShadow (dismissed):

MegaIng (dismissed):

EnderShadow:
I still think that the Q bit should be undefined behavior instead of illegal in the cases we've made it illegal, but that's just a personal preference.

MegaIng (approved):

PR #13, opened 2022-05-30
Remove exten control register and add second cpuid control register

EnderShadow:
This PR accomplishes a few things:

1. The EXTEN CR is removed since it's not actually used by any currently defined instructions.
    - Only a few planned extensions use it, and at least 2 of those require a partially enabled state in the first place.
    - This CR will be added back in a more useful way by extensions that use it when they are drafted.
2. A second CPUID CR is added.
    - This allows for a more feature rich base for etca
    - This also allows for more extensions that don't require the 32 or 64 bit operation extensions just to check if they're present

### DIFF
--- a/base-isa.md
+++ b/base-isa.md
@@ -97,11 +97,11 @@ TODO: This is a baseline, very much still floating
 
 Undefined control registers are reserved and reading from or writing to them is undefined behavior.
 
-| `CRN`  | NAME       | Description                                                                                                                            | Comment |
-|--------|------------|----------------------------------------------------------------------------------------------------------------------------------------|---------|
-| `0000` | `CPUID1`   | Reading from this control register puts the first set of available CPU extensions in the destination register. Writing to it is a NOP  | (1)     |
-| `0001` | `CPUID2`   | Reading from this control register puts the second set of available CPU extensions in the destination register. Writing to it is a NOP | (1)     |
-| `0010` | `FEAT`     | Reading from this control register puts the available CPU features in the destination register. Writing to it is a NOP                 | (2)     |
+| `CRN`  | NAME       | Description                                                                                                                             | Comment |
+|--------|------------|-----------------------------------------------------------------------------------------------------------------------------------------|---------|
+| `0000` | `CPUID1`   | Reading from this control register puts the first set of available CPU extensions in the destination register. Writing to it is a NOP.  | (1)     |
+| `0001` | `CPUID2`   | Reading from this control register puts the second set of available CPU extensions in the destination register. Writing to it is a NOP. | (1)     |
+| `0010` | `FEAT`     | Reading from this control register puts the available CPU features in the destination register. Writing to it is a NOP.                 | (2)     |
 
 1) CPUID uses a bitfield to specify the available extensions. A value of 0 means no extensions are available.
 2) FEAT uses a bitfield to specify the available features. A value of 0 means no features are available.
--- a/base-isa.md
+++ b/base-isa.md
@@ -97,11 +97,11 @@ TODO: This is a baseline, very much still floating
 
 Undefined control registers are reserved and reading from or writing to them is undefined behavior.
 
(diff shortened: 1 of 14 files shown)


extensions/arbitrary-stack-pointer/README.md
```diff
@@ -3,7 +3,7 @@
 **Extension State: Under Development**  
 **Enabled by Default: Yes**  
 **Requires: Base**  
-**CPUID Bit: 8**
+**CPUID 1 Bit: 7**
```
MegaIng: Minor thing, I would this to be `CPUID1`, so that it's consistent everywhere

base-isa.md
```diff
@@ -99,13 +99,12 @@ Undefined control registers are reserved and reading from or writing to them is
 
 | `CRN`  | NAME       | Description                                                                                                              | Comment |
 |--------|------------|--------------------------------------------------------------------------------------------------------------------------|---------|
-| `0000` | `CPUID`    | Reading from this control register puts the available CPU extensions in the destination register. Writing to it is a NOP | (1)     |
-| `0001` | `EXTEN`    | This control register specifies which available extensions are enabled or disabled.                                      | (2)     |
-| `0010` | `FEAT`     | Reading from this control register puts the available CPU features in the destination register. Writing to it is a NOP   | (3)     |
+| `0000` | `CPUID1`   | Reading from this control register puts the available CPU extensions in the destination register. Writing to it is a NOP | (1)     |
+| `0001` | `CPUID2`   | Reading from this control register puts the available CPU extensions in the destination register. Writing to it is a NOP | (1)     |
```
MegaIng: This sounds like it's a full duplicate of `CPUID1`.

EnderShadow: I wasn't sure how to properly word it, but it's a second CPUID CR which has a different set of extensions specified. The main purpose is to allow us to have more extensions which don't depend on 32/64 bit operations to detect their presence or lack thereof. It also gives us a second set of easy to access extensions which would be useful to detect.

extensions/arbitrary-stack-pointer/README.md
```diff
@@ -3,7 +3,7 @@
 **Extension State: Under Development**  
 **Enabled by Default: Yes**  
 **Requires: Base**  
-**CPUID Bit: 8**
+**CPUID1 Bit: 7**
```
JKTKops: I think instead of this we could consider CPUID as a large struct spanning more than one register (this is typical for such operations in real ISAs) so the notation for this would be

> **CPUID Bit: CP1.7**

Or something to that effect. What do you think?

rationale.md
```diff
@@ -78,14 +78,19 @@ Of course, an extension will indicate that the processor supports executable RAM
...
+          mov  %r0, cpuid2
           and  %r0, 0xF
           cmp  %r0, 0xF     ; are the core extensions available?
           jeq  can_work
+fail:
           hlt               ; halt and catch fire
 can_work: ... ; rest of the program (including interupt setup)
 ```
-These four extensions together make up `ETC.a.F` or the "core extensions".
+These eight extensions together make up `etca.f.f.0` or the "core extensions".
```
JKTKops:
```suggestion
These eight extensions together make up `ETCa.F.F.0` or the "core extensions".
```

JKTKops (dismissed):

extensions/arbitrary-stack-pointer/README.md
EnderShadow: That's always an option. That's what I was going for with how I specified dependencies in the extensions table.

rationale.md
EnderShadow: This was something I wasn't sure of. Do we want 'ETCa' or 'ETCA'?

extensions/arbitrary-stack-pointer/README.md
EnderShadow: Should be resolved now.

MegaIng (approved):

PR #21, opened 2022-07-06
Update expanded registers

### LINKED ISSUE #20: calling convention for sN registers of REX
Commit 7f66dbe updates the documentation of the [Expanded Registers](https://github.com/ETC-A/etca-spec/blob/main/extensions/expanded-registers) extension, to the current names for the added registers. I think its "who saves it" column may need an update to match.

Formerly, the registers r13-r15 had aliases a8, t0, and t1, and correspondingly were listed as call**er**-saved. Register aliases s2-s4 were on registers r24-r26, and were call**ee**-saved. Now r13-r15 _are_ s2-s4, but they've kept their call**er**-saved recommendation, which seems inconsistent with the caption text that "sN registers are registers that must be saved before re-using."

Should they be updated to call**ee**-saved, like the other sN registers? Or is it intended that these registers keep the call**er**-saved status?

Ezra:
Resolves #20 and makes some stylistic edits to the surrounding text.

### DIFF
--- a/extensions/expanded-registers/README.md
+++ b/extensions/expanded-registers/README.md
@@ -9,6 +9,20 @@
 
 This extension adds an instruction prefix byte expanding the number of available registers from 8 registers to 16, and suggests an extension of the [Stack & Functions](../stack-and-functions/readme.md) extension's calling conventions to the added registers.
 
+# Added Instructions
+
+The following 1-byte instruction prefix is now defined.
+
+| `REX`  |     |     |     |     |
+|:-------|-----|-----|-----|-----|
+| `1100` | `Q` | `A` | `B` | `X` |
+
+When used with an instruction that uses an `A` register, `REX.A` provides a 4th bit to be used as the most significant bit of the register specifier.
+When used with an instruction that uses a `B` register (including an `SIB.B` if [MO1](../memory-operands-1/README.md) or [MO2](../memory-operands-2/README.md) is available), `REX.B` provides a 4th bit for the specifier in the same way.
+When used with an instruction that uses an `SIB.X` register, `REX.X` provides a 4th bit for the specifier in the same way.
+
+`REX.Q` is reserved.
+
 # Calling Conventions
 
 When used together with the [Stack & Functions](../stack-and-functions/readme.md) extension, the following table gives the recommended interpretation of the expanded registers.
@@ -28,17 +42,3 @@ When used together with the [Stack & Functions](../stack-and-functions/readme.md
 - tN registers are temporary registers which a function can use however it wants.
 
 Remember that the above are just suggestions. Until a more complete specification of a standard calling convention for ETC-A is specified, expect some implementations not to follow them. 
-
-# Added Instructions
-
-The following 1-byte instruction prefix is now defined.
-
-| `REX`  |     |     |     |     |
-|:-------|-----|-----|-----|-----|
-| `1100` | `Q` | `A` | `B` | `X` |
-
-When used with an instruction that uses an `A` register, `REX.A` provides a 4th bit to be used as the most significant bit of the register specifier.
-When used with an instruction that uses a `B` register (including an `SIB.B` if [MO1](../memory-operands-1/README.md) or [MO2](../memory-operands-2/README.md) is available), `REX.B` provides a 4th bit for the specifier in the same way.
-When used with an instruction that uses an `SIB.X` register, `REX.X` provides a 4th bit for the specifier in the same way.
-
-`REX.Q` is reserved.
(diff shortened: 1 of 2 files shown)


EnderShadow (changes requested):
I prefer that we stick with "caller saved" and "callee saved" since that has a technical meaning whereas "call-clobbered" and "call-preserved" don't have the same meaning.

JKTKops:
Indeed, caller saved and callee saved are technical terms. The confusion is unfortunate but they are standard now.

extensions/stack-and-functions/README.md
```diff
@@ -9,24 +9,24 @@
...
-- tN registers are temporary registers which a function can use however it wants.
+- vN registers are the first N return values from a function call. Additional return values should be pushed to the stack.
+- sN registers are registers that must be saved before, and restored after, use.
+- tN registers (see the [Expanded Registers](../expanded-registers/README.md) extension) are temporary registers which a function can use however it wants.
 - bp is the base pointer register and stores the address of the bottom of the stack for this function.
-- sp is the stack pointer register and stores the address of the top of the stack for this function. This is hardcoded into the push and pop instructions
-- ln is the link register and stores the address that this function should return to. This is hard coded into the call and return instructions
+- sp is the stack pointer register and stores the address of the top of the stack for this function. This is hardcoded into the push and pop instructions.
+- ln is the link register and stores the address that this function should return to. This is hard coded into the call and return instructions.
```
JKTKops:
```suggestion
- ln is the link register and stores the address that this function should return to. This is hardcoded into the call and return instructions.
```

extensions/stack-and-functions/README.md
```diff
@@ -9,24 +9,24 @@
...
+| Register | Alias     | Calling Convention |
+|----------|-----------|----------------|
+| `r0`     | `a0`/`v0` | call-clobbered |
+| `r1`     | `a1`/`v1` | call-clobbered |
+| `r2`     | `a2`      | call-clobbered |
+| `r3`     | `s0`      | call-preserved |
+| `r4`     | `s1`      | call-preserved |
+| `r5`     | `bp`      | call-preserved |
+| `r6`     | `sp`      | call-preserved |
+| `r7`     | `ln`      | call-clobbered |
```
JKTKops:
```suggestion
| Register | Alias     | Calling Convention |
|----------|-----------|----------------|
| `r0`     | `a0`/`v0` | caller saved |
| `r1`     | `a1`/`v1` | caller saved |
| `r2`     | `a2`      | caller saved |
| `r3`     | `s0`      | callee saved |
| `r4`     | `s1`      | callee saved |
| `r5`     | `bp`      | callee saved |
| `r6`     | `sp`      | callee saved |
| `r7`     | `ln`      | caller saved |
```
Capitalization doesn't matter much to me. They were capitalized before but leaving them lowercase is fine.

extensions/expanded-registers/README.md
```diff
@@ -40,3 +23,22 @@ When used with an instruction that uses an `SIB.X` register, `REX.X` provides a
...
+| Register | Alias | Calling Convention |
+|----------|-------|----------------|
+| `r8`     | `t0`  | call-clobbered |
+| `r9`     | `t1`  | call-clobbered |
+| `r10`    | `t2`  | call-clobbered |
+| `r11`    | `t3`  | call-clobbered |
+| `r12`    | `t4`  | call-clobbered |
+| `r13`    | `s2`  | call-preserved |
+| `r14`    | `s3`  | call-preserved |
+| `r15`    | `s4`  | call-preserved |
```
JKTKops, line 39: Same as suggested change below; use technical terms.

extensions/expanded-registers/README.md
```diff
@@ -7,31 +7,14 @@
...
+This extension adds an instruction prefix byte expanding the number of available registers from 8 registers to 16, and suggests an extension of the [Stack & Functions](../stack-and-functions/readme.md) extension's calling conventions to the added registers.
 
 # Added Instructions
 
 The following 1-byte instruction prefix is now defined.
 
-| `REX`   | | | | |
-|:-------------|---|----|----|----|
+| `REX`  |     |     |     |     |
+|:-------|-----|-----|-----|-----|
```
JKTKops, line 17: Cleaning these up is fine, but remember that markdown doesn't actually care about the spacing and spacing things "correctly" gets messed up if the table is ever modified. This has come up in our tables before.

JKTKops (changes requested):

EnderShadow:
As this has been idle for a bit, I'm closing it and putting the changes in #24

PR #22, opened 2022-07-14
Add interrupt extension

EnderShadow:


extensions/interrupts/README.md
```diff
@@ -0,0 +1,82 @@
...
+| `IRET` | `00 10 1111`  | `0000 0000` | Returns from the current interrupt. Executing this when not in an interrupt causes a General Protection Fault.                                       |
+|        | `00 1? 1111`  | `???? ????` | When `? ???? ????` is not `0 0000 0000`, reserved for future extensions.                                                                             |
+
+# Interrupt Flow
+
+CPU interrupts from external hardware are rising edge triggered.
+
+If a synchronous interrupt occurs while the CPU is already handling another interrupt or the corresponding bit in the `INT_MASK` CR is unset, the CPU _must_ reset as there is no way for it to handle the interrupt.
+If the CPU is not handling an interrupt when a synchronous interrupt occurs and the corresponding mask bit is set, then the corresponding bit in the `INT_PENDING` CR will be set and the CPU will immediately
+transition to handling that interrupt without finishing it's current instruction. This means that the CPU _must_ be in the same effective state as before execution of the instruction was attempted aside from what is
```
JKTKops:
```suggestion
transition to handling that interrupt without finishing its current instruction. This means that the CPU _must_ be in the same effective state as before execution of the instruction was attempted aside from what is
```

JKTKops (dismissed):
LGTM for in-development extension. Also noticed a typo

extensions/interrupts/README.md
EnderShadow: For some reason github wouldn't let me add that suggestion from my end so I fixed it with another commit.

JKTKops (approved):

PR #25, opened 2022-07-18
Bit Manipulation 1

EnderShadow:


MegaIng (dismissed):
For all operations the behavior when `B>M` is should be defined, especially for the rotate instructions. I am not really sure what the correct behavior should be, but currently `ROL(A=9,B=15, M=7)` returns `4`, which I am sure shouldn't be the answer (unless it's explicitly undefined behavior).

Also, I assume `B` is always taken as an unsigned argument? And this should also effect the 5bit immediate and mentioned I would think.

EnderShadow:
That rotate would not be 4. I forgot to mention it in the spec, but only the lower `b` bits will be used where `b = S + 3` This makes the above `ROL(A=9, B=15, SS=0)` return `132` or `-124`.

Also, the expression for arithmetic right shift doesn't properly handle `B = 0`.

MegaIng (approved):

PR #27, opened 2022-07-19
Modify spec to allow for the overlap between SLO/POP, READCR/LEA, and WRITECR/INT/IRET

EnderShadow:


extensions/full-immediates/README.md
```diff
@@ -17,7 +17,8 @@ In Base ETCa, the second byte, or "ABM byte" takes this format:
 +-----+-----+----+
 ```
 
-Only `MM=00` is defined by base. This extension applies to any instruction which uses an ABM byte.
+Only `MM=00` is defined by base. This extension applies to any instruction which uses an ABM byte. This extension also allows the use of register-register mode for the purpose of
```
JKTKops:
```suggestion
Only `MM=00` is defined by base. This extension applies to any instruction which uses an ABM byte. This extension also allows the use of register-register format for the purpose of
```

This is very important - the instructions do not suddenly gain a register-register mode of operation. They hijack the register-register _format_ for the means of getting encoding space which the register-register format cannot use anyway.

extensions/full-immediates/README.md
```diff
@@ -17,7 +17,8 @@ In Base ETCa, the second byte, or "ABM byte" takes this format:
...
+full immediates on instructions which are immediate only. Examples of such instructions are `SLO`, `READCR`, and `WRITECR`.
```
JKTKops, line 21:
```suggestion
full immediates on instructions which are immediate only. Examples of such instructions are `SLO`, `READCR`, and `WRITECR`.
Note that the corresponding operations in the register-register format are `POP`, `LEA`, and `INT`/`IRET`, for which immediate operands do not make sense.

When such an opcode appears in the register-register format, in a full immediate operand mode, it is treated
by this spec exactly as if it were the register-immediate format opcode.
```

I'm not sure off the top of my head what the last one is so fix that :P

extensions/memory-operands-1/README.md
```diff
@@ -114,7 +114,7 @@ as though the operand values were the size specified by the instruction's Operan
 | LEA r, m | `00SS1110` | `ABM` | Loads the address specified by the second operand into the register specified by the first. Note that the address _itself_ is stored in the register, not the contents of memory at that address. The result stored in the register must respect the operand size attribute. The computed address must respect the address mode.
 
 `LEA` means "load effective address." If the first operand is not a register, or the second is not a memory location, the instruction must be treated as an
-undefined instruction. Such cases are **not** reserved - they are explicitly undefined.
+undefined instruction. Such cases are **not** reserved - they are explicitly undefined. The exception to this is the full immediate mode of this instruction. That is a `READCR` instruction and NOT an `LEA` instruction.
```
JKTKops:
```suggestion
undefined instruction. Such cases are **not** reserved - they are explicitly undefined. Note that as per the
[Full Immediates extension](../full-immediates/README.md) in the specific case of `MM=01, ABM.regB=01x`
(the "full immediate" operand modes) this opcode corresponds to `READCR`, not `LEA`.
```

This should be applied to the MO2 document as well.

JKTKops (changes requested):

extensions/full-immediates/README.md
EnderShadow: The last one was INT and IRET, but I already modified the interrupt spec in a different PR to fix the issue with it so it's not conflicting in the first place.

extensions/memory-operands-2/README.md
```diff
@@ -57,7 +57,7 @@ In all other instruction formats, `REX.Q` remains reserved.
 | LEA r, m | `00SS1110` | `ABM` | Loads the address specified by the second operand into the register specified by the first. Note that the address _itself_ is stored in the register, not the contents of memory at that address. The result stored in the register must respect the operand size attribute. The computed address must respect the address mode.
 
 `LEA` means "load effective address." If the first operand is not a register, or the second is not a memory location, the instruction must be treated as an
-undefined instruction. Such cases are **not** reserved - they are explicitly undefined.
+undefined instruction. Such cases are **not** reserved - they are explicitly undefined. The exception to this is the full immediate mode of this instruction. That is a `READCR` instruction and NOT an `LEA` instruction.
```
EnderShadow:
```suggestion
undefined instruction. Such cases are **not** reserved - they are explicitly undefined. Note that as per the
[Full Immediates extension](../full-immediates/README.md) in the specific case of `MM=01, ABM.regB=01x`
(the "full immediate" operand modes) this opcode corresponds to `READCR`, not `LEA`.
```

JKTKops (approved):

PR #33, opened 2022-07-23
Modify feature description

EnderShadow:


NotNotQuinn (approved):
I like this description a lot better, its a lot more clear to me what should be in the "feature" category vs "extension". LGTM!

JKTKops (approved):

EnderShadow:
> I like this description a lot better, its a lot more clear to me what should be in the "feature" category vs "extension". LGTM!

Thanks, unfortunately your approvals don't count towards approval vs rejection since you're not one of the maintainers of the repo. Thanks for the approval and input regardless though!

PR #34, opened 2022-07-24
Specify mapping for flags in the `FLAGS` CR

EnderShadow:


NotNotQuinn:
Is it possible to specify the last 4 bits of the `FLAGS` CR to be "reserved for extensions or another purpose". It seems a bit wasteful to leave these bits unspecified, unless more ALU flags are planned to be added? (I dont know)

EnderShadow:
Since they aren't specified, they are implicitly reserved. I'll explicitly reserve them tomorrow.

MegaIng (approved):

PR #36, opened 2022-07-26
Clarify prefix usage

EnderShadow:


extensions/variable-width-instruction-standard.md
```diff
@@ -21,6 +21,8 @@ Instruction prefixes are a sequence of one or more bytes which modify the instru
 1. Conditional Prefix
 2. Expanded Registers
 
+Unless specified otherwise, each prefix can only be used _at most_ once per instruction
```
NotNotQuinn: Oops
```suggestion
Unless specified otherwise, each prefix can only be used _at most_ once per instruction.
````
Otherwise I like these changes, I feel like they were implied but its always better to have it documented.

MegaIng (approved):

PR #37, opened 2022-07-26
Create notation reference.

NotNotQuinn:
I'll be creating a reference for notation as I explore the ETCa project. This is for me and hopefully anyone who will be looking into the project in the future. This is going to be mostly intended for people just starting to look into ETCa who havent been participating as the notation evolved, but also a handy reference if you forget.

So far I plan to explain the following things (and note where they are introduced in the spec) + more:
 - [x] That the notation reference is NOT part of the ETCa spec.
 - [ ] How to understand and read `ETCa.0.0.0` and `ETCa.F.F.0`, etc.
 - [ ] What a `SIB` byte is.
 - [ ] What an `ABM` byte is.
 - [ ] What a `REX` byte is, and `REX.Q`, `REX.A`, etc.
 - [ ] Other common abbreviations, such as `CR`, that are just implied but never explicitly explained.
 - [ ] Some examples of how the instructions can be laid out byte by byte.

Currently looking for more notations to add, and always looking for feedback on accuracy as I am new to the project. So feel free to edit this description to add more things to describe!

### DIFF
--- a/notation.md
+++ b/notation.md
@@ -1,0 +1,6 @@
+> ###### Preamble
+> This file is not part of any formal specification of any kind, but is more indended to be used
+> be used as a documentation of the notations used throughout the ETCa project, as they are 
+> specific to the ETCa architechture and community, and sometimes impossible to decipher without prior knowledge.
+
+# General Notation
--- a/notation.md
+++ b/notation.md
@@ -1,6 +1,7 @@
 > ###### Preamble
 > This file is not part of any formal specification of any kind, but is more indended to be used
-> be used as a documentation of the notations used throughout the ETCa project, as they are 
-> specific to the ETCa architechture and community, and sometimes impossible to decipher without prior knowledge.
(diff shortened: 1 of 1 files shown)


EnderShadow:
Nice! I do want to mention that the interrupt extension requires a von-neuman architecture though. So I think it would be `ETCa.F.F.1`

NotNotQuinn:
Brief schedule update:
Currently I'm planning on doing the following things, along with whatever IRL things come up:

- ~~Finish my 7 segment display for TC~~
- ~~Finish the TC campaign~~
- Build an `ETCa.0.0.0` machine in TC
- Work out a group of extensions that I would like to implement
- Implement them

So this won't likely be worked on until I have finished the TC campaign.

EnderShadow:
Are you still working on this @NotNotQuinn?

NotNotQuinn:
I would say probably not. Maybe at some point, but I think this can be closed for now.

PR #38, opened 2022-08-04
Put carry flag in bit 0

EnderShadow:
This makes implementation of ADC, SBC, and RSBC easier by allowing the flags to be directly input to the carry in of the ALU of TC


EnderShadow:
Nevermind. I forgot that bit indexers were a thing.

PR #39, opened 2022-08-13
Ensure consistency in the entire spec

EnderShadow:
This also adds some missing parts of the spec.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,21 @@
-# ETC.A
+# ETCa
 Extensible Turing Complete Architecture
 
-This repo is for designing an ISA that has a simple core and a lot of extensions that can be added ontop.
-It's primary goal is to be build in the game [Turing Complete](https://turingcomplete.game) and to be easy to implement in there.
+This repo is for designing an ISA that has a simple base and multiple extensions that can be added ontop. It's primary goal is to be easy to build in the game
+[Turing Complete](https://turingcomplete.game) after the second architecture is completed. Several other goals include being educational and being practical for a real CPU.
 
-An extensible assembler built for this arch is [here](https://github.com/MegaIng/etca-asm)
+An extensible assembler built for this architecture is [here](https://github.com/ETC-A/etca-asm)
+
+## Terminology
+
+_Must_: Specifies that a specific behavior is required by the specification.
+
+_Should_: Specifies that a specific behavior is recommended but not required.
+
+_May_: Specifies that a specific behavior is allowed and not encouraged or discouraged.
+
+_Illegal_: Specifies that a specific behavior or instruction is **NOT** allowed in a compliant implementation and must be handled as specified by the ISA.
+
+_Reserved_: Specifies that the behavior or instruction will be specified in a future extension or feature and should be treated as _illegal_ until it is both specified and implemented.
+
+_Unspecified_: Specifies that the behavior or instruction is intentionally unconstrained and may be freely handled any way an implementation chooses.
--- a/README.md
+++ b/README.md
@@ -18,4 +18,5 @@ _Illegal_: Specifies that a specific behavior or instruction is **NOT** allowed 
 
 _Reserved_: Specifies that the behavior or instruction will be specified in a future extension or feature and should be treated as _illegal_ until it is both specified and implemented.
 
-_Unspecified_: Specifies that the behavior or instruction is intentionally unconstrained and may be freely handled any way an implementation chooses.
+_Unspecified_: Specifies that the behavior or instruction is intentionally unconstrained and may be freely handled any way an implementation chooses. Extensions and features can specify
+or restrict _unspecified_ behavior.
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # ETCa
 Extensible Turing Complete Architecture
 
-This repo is for designing an ISA that has a simple base and multiple extensions that can be added ontop. It's primary goal is to be easy to build in the game
+This repo is for designing an ISA that has a simple base and multiple extensions that can be added on top. Its primary goal is to be easy to build in the game
 [Turing Complete](https://turingcomplete.game) after the second architecture is completed. Several other goals include being educational and being practical for a real CPU.
 
-An extensible assembler built for this architecture is [here](https://github.com/ETC-A/etca-asm)
+An extensible assembler built for this ISA is [the etca-asm repository](https://github.com/ETC-A/etca-asm) in this organization.
 
 ## Terminology
 
--- a/README.md
+++ b/README.md
@@ -8,11 +8,7 @@ An extensible assembler built for this ISA is [the etca-asm repository](https://
 
 ## Terminology
 
-_Must_: Specifies that a specific behavior is required by the specification.
-
-_Should_: Specifies that a specific behavior is recommended but not required.
-
-_May_: Specifies that a specific behavior is allowed and not encouraged or discouraged.
+See [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt) for some of the terminology used in the documentation for ETCa.
 
 _Illegal_: Specifies that a specific behavior or instruction is **NOT** allowed in a compliant implementation and must be handled as specified by the ISA.
 
--- a/base-isa.md
+++ b/base-isa.md
@@ -9,7 +9,7 @@
 
 # Base Instructions
 
-This is a group of 16bit fixed width instructions. The motivation for the chosen instructions is to be both complete and easy to build in game.
+This is a group of 16 bit fixed width instructions. The motivation for the chosen instructions is to be both complete and easy to build in game.
 
 The highest two bits of the first byte are a format marker:
 
@@ -50,15 +50,15 @@ Execution of an _illegal_ or _reserved_ instruction is _unspecified_ behavior. A
 Both computation formats share a lot of similarities. For both, the first byte has the structure `0x SS CCCC`.
 
 - `SS` is a size marker and reserved for extensions. For the base instructions this _must_ always be `01`
-- `CCCC` is a 4bit opcode deciding which operation to execute.
+- `CCCC` is a 4 bit opcode deciding which operation to execute.
 
 ### 2 Register Computation
 
-The second byte has the format `AAA BBB MM` where `AAA` and `BBB` are references to registers. `AAA` is the destination register and left source register. `BBB` is the right source register. The only exception is the RSUB instruction where `AAA` and `BBB` are swapped for the purpose of being source registers. `MM` is the memory mode which will be defined in an extension and _must_ be `00`
+The second byte has the format `AAA BBB MM` where `AAA` and `BBB` are references to registers. `AAA` is the destination register and left source register. `BBB` is the right source register. There are 2 exceptions to this. The first is the RSUB instruction where `AAA` and `BBB` are swapped for the purpose of being source registers. The second exception is that the `B` argument specifies which memory address or control register to write to. `MM` is the memory mode which will be defined in an extension and is only defined for `00` in the base isa.
 
 ### Immediate Computation
 
-The second byte has the format `RRR IIIII`, where the 5bit immediate acts as operand `B`. The immediate is sign extended for operations 0-7 and operation 9. The immediate is zero extended for operation 8 and operations 10-15.
+The second byte has the format `RRR IIIII`, where the 5 bit immediate acts as operand `B`. The immediate is sign extended for operations 0-7 and operation 9. The immediate is zero extended for operation 8 and operations 10-15.
 
 ### Opcodes
 
@@ -123,7 +123,7 @@ Aside from program rom starting at address 0x8000, there are no requirements for
 
 ## Jump Instructions
 
-Here the first byte has the format `10 0 D CCCC` where `D` fills the high byte of the displacement. `CCCC` is the condition to check. The second byte is a single 8bit immediate representing the low byte of the displacement. This combined 9 bit displacement (sign extended to the address width which is 16 bits in the base isa) is added to the base address of the current instruction and stored in the program counter.
+Here the first byte has the format `10 0 D CCCC` where `D` fills the high byte of the displacement. `CCCC` is the condition to check. The second byte is a single 8 bit immediate representing the low byte of the displacement. This combined 9 bit displacement (sign extended to the address width which is 16 bits in the base isa) is added to the base address of the current instruction and stored in the program counter.
 
 ### Conditions
 
--- a/base-isa.md
+++ b/base-isa.md
@@ -5,11 +5,11 @@
 - 8 16 bit registers
 - 4 flags set upon (some) ALU operations: Zero (`Z`), Negative (`N`), Carry (`C`), Overflow (`V`)
 - CPU Execution starts at address 0x8000
-  - This is only relevant if one or more extension are implemented which expose the instruction pointer in some way.
+  - This is not relevant for the ISA with no extensions.
 
 # Base Instructions
 
-This is a group of 16 bit fixed width instructions. The motivation for the chosen instructions is to be both complete and easy to build in game.
+This is a group of 16 bit fixed width instructions.
 
 The highest two bits of the first byte are a format marker:
 
@@ -28,7 +28,7 @@ The highest two bits of the first byte are a format marker:
 | `01 01 CCCC`  | `RRR IIIII`  | immediate and 1 register computation     |
 | `0? SS CCCC`  | `???? ????`  | when `SS != 01`, reserved for extensions |
 | `10 0 D CCCC` | `DDDDDDDD`   | (conditional) relative jump instruction  |
-| `10 1 ?????`  | `?????????`  | reserved for extension                   |
+| `10 1 ?????`  | `?????????`  | reserved for extensions                  |
 | `11 ??????`   | `?????????`  | reserved for extensions                  |
 
 | Symbol | Meaning         |
@@ -38,8 +38,10 @@ The highest two bits of the first byte are a format marker:
 | I      | Immediate       |
 | D      | Displacement    |
 | S      | Operation Size  |
-| M      | Memory Mode     |
+| M      | Memory Addressing Mode     |
 | ?      | Arbitrary Value |
+
+Notice that whenever `MM` bits are present, it is reserved for their value to be anything other than `00`. Similarly, it is reserved for `SS` bits to have values other than `01`.
 
 ### Execution of Illegal and Reserved Instructions
 
@@ -49,7 +51,7 @@ Execution of an _illegal_ or _reserved_ instruction is _unspecified_ behavior. A
 
 Both computation formats share a lot of similarities. For both, the first byte has the structure `0x SS CCCC`.
 
-- `SS` is a size marker and reserved for extensions. For the base instructions this _must_ always be `01`
+- `SS` is a size marker and reserved for extensions. For the base instructions this is always `01`.
 - `CCCC` is a 4 bit opcode deciding which operation to execute.
 
 ### 2 Register Computation
@@ -85,7 +87,7 @@ The second byte has the format `RRR IIIII`, where the 5 bit immediate acts as op
 1) Enables NEG and NOT to be encoded as `RSUB r, imm`.
 2) Placed here to ease decoding; `xx11` => do not store result.
 3) Designed to allow for building a larger immediate value. To reach a full 16 bit immediate, a 4th `SLO` or an additional `NOT` may be required.
-4) Control registers can only be accessed with immediates as a way to prevent potential security vulnerabilities.                                                                  
+4) Control registers can only be accessed with immediates.
 5) The C and V flags are in an _unspecified_ state after execution of these instructions. Extensions _may_ mandate a particular behavior, with good enough reason, but must **NOT** mandate that the value of these flags after the operation depends on their value before the operation.
 6) These instructions do not have a 2 register mode. The corresponding bit patterns (`00 SS 11??`) for the first byte are _reserved_. This can easily be detected by using a similar method to 2).
 7) This instruction zero extends argument B as if the value read is of the size specified by the SS bits in the instruction (assuming a relevant extension is present).
@@ -112,18 +114,18 @@ All IO is memory mapped
 
 ### Separation of Program ROM and RAM
 
-Aside from program rom starting at address 0x8000, there are no requirements for how the memory address space is layed out. For ease of implementation, this spec does **NOT** require program memory to be accessible with the load and store instructions nor does it require that RAM be executable. Loads and stores of program memory are _unspecified_ behavior. Executing from RAM is _unspecified_ behavior. Future extensions and/or features will change this behavior
+Aside from program rom starting at address 0x8000, there are no requirements for how the memory address space is layed out. For ease of implementation, this spec does **NOT** require program memory to be accessible with the load and store instructions nor does it require that RAM be executable. Loads and stores of program memory are _unspecified_ behavior. Executing from RAM is _unspecified_ behavior. Future extensions and/or features will change this behavior.
 
 ## Flag Semantics
 
 1) The `Z` flag indicates that the result was zero.
 2) The `N` flag indicates that the result, interpreted as a 2's complement number, was negative.
-3) The `C` flag describes whether an additive instruction caused a carry, or a subtractive computation caused a borrow. The concept of "borrow" is exactly the same as in long-hand decimal subtraction the way you might do it on paper. The flag is set if the subtraction would borrow out of the next most significant bit. A borrow happens precisely when the corresponding addition does _not_ carry. In other words, in order to subtract, you (a) complement the B input, (b) set the carry in, (c) set C to the inverse of the carryout.
+3) The `C` flag describes whether an additive instruction caused a carry, or a subtractive computation caused a borrow. The concept of "borrow" is exactly the same as in long-hand decimal subtraction the way you might do it on paper. The flag is set if the subtraction would borrow out of the next most significant bit. A borrow happens precisely when the corresponding 2's-complement addition does _not_ carry. In other words, in order to subtract, you (a) complement the B input, (b) set the carry-in, (c) set C to the inverse of the carryout.
 4) The `V` flag indicates that the operation caused overflow. This means that a loss of precision occurred for a signed integer calculation (eg. positive + positive = negative or negative + negative = positive).
 
 ## Jump Instructions
 
-Here the first byte has the format `10 0 D CCCC` where `D` fills the high byte of the displacement. `CCCC` is the condition to check. The second byte is a single 8 bit immediate representing the low byte of the displacement. This combined 9 bit displacement (sign extended to the address width which is 16 bits in the base isa) is added to the base address of the current instruction and stored in the program counter.
+Here the first byte has the format `10 0 D CCCC` where `D` fills the high byte of the displacement. `CCCC` is the condition to check. The second byte is a single 8 bit immediate representing the low byte of the displacement. This combined 9 bit displacement (sign extended to the address width, which is 16 bits in the base isa) is added to the base address of the current instruction and stored in the program counter.
 
 ### Conditions
 
--- a/base-isa.md
+++ b/base-isa.md
@@ -20,16 +20,16 @@ The highest two bits of the first byte are a format marker:
 
 # Overview
 
-| First byte    | Second Byte  | Comment                                  |
-|:--------------|:-------------|:-----------------------------------------|
-| `00 01 CCCC`  | `RRR RRR 00` | 2 register computation                   |
-| `00 01 11??`  | `???? ????`  | reserved for extensions                  |
-| `00 SS CCCC`  | `RRR RRR MM` | when `MM != 00`, reserved for extensions |
-| `01 01 CCCC`  | `RRR IIIII`  | immediate and 1 register computation     |
-| `0? SS CCCC`  | `???? ????`  | when `SS != 01`, reserved for extensions |
-| `10 0 D CCCC` | `DDDDDDDD`   | (conditional) relative jump instruction  |
-| `10 1 ?????`  | `?????????`  | reserved for extensions                  |
-| `11 ??????`   | `?????????`  | reserved for extensions                  |
+| First byte    | Second Byte  | Comment                                          |
+|:--------------|:-------------|:-------------------------------------------------|
+| `00 01 CCCC`  | `RRR RRR 00` | 2 register computation                           |
+| `00 SS CCCC`  | `RRR RRR MM` | when not matching above, reserved for extensions |
+| `00 01 11??`  | `???? ????`  | reserved for extensions                          |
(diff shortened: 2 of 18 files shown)


README.md
```diff
@@ -1,7 +1,21 @@
...
+
+_Should_: Specifies that a specific behavior is recommended but not required.
+
+_May_: Specifies that a specific behavior is allowed and not encouraged or discouraged.
+
+_Illegal_: Specifies that a specific behavior or instruction is **NOT** allowed in a compliant implementation and must be handled as specified by the ISA.
+
+_Reserved_: Specifies that the behavior or instruction will be specified in a future extension or feature and should be treated as _illegal_ until it is both specified and implemented.
+
+_Unspecified_: Specifies that the behavior or instruction is intentionally unconstrained and may be freely handled any way an implementation chooses.
```
MegaIng: Maybe it should be mentioned here that features could potentially specified previously unspecified behavior?

base-isa.md
```diff
@@ -1,14 +1,15 @@
 # General design
 
-- 16bit word size
-- 8 16bit registers
+- 16 bit word size
```
MegaIng, line 3: We should decide if we want to keep `16bit` or `16 bit`. I personally prefer the former. You changed it here into the latter, but missed it a few lines down.

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
...
 
-- `SS` is a size marker and reserved for extensions. For the base instructions this should always be `01`
-- `CCCC` is a 4bit opcode deciding which operation to execute. This can be an ALU or a memory load/store instruction
+- `SS` is a size marker and reserved for extensions. For the base instructions this _must_ always be `01`
+- `CCCC` is a 4bit opcode deciding which operation to execute.
 
 ### 2 Register Computation
 
-The second byte has the format `AAA BBB 00` where `AAA` and `BBB` are references to registers. `AAA` is the destination register and left source register. `BBB` is the right source register. The only exception is the RSUB instruction where `AAA` and `BBB` are swapped for the purpose of being source registers.
+The second byte has the format `AAA BBB MM` where `AAA` and `BBB` are references to registers. `AAA` is the destination register and left source register. `BBB` is the right source register. The only exception is the RSUB instruction where `AAA` and `BBB` are swapped for the purpose of being source registers. `MM` is the memory mode which will be defined in an extension and _must_ be `00`
```
MegaIng: Another exception are `store` and `writecr` where `BBB` more or less gives the destination. (a few people on discord missed that before).

Also, the last sentence should maybe have a clause `must be 00 for base`?

MegaIng (changes requested):

README.md
EnderShadow: Good catch. Fixed.

base-isa.md
EnderShadow: I fixed it so that the spec is consistent in regards to this.

base-isa.md
EnderShadow: Fixed.

MegaIng (dismissed):
Only that few uses of ``N bit``? I am suprised by that tbh. I didn't see anything else.

EnderShadow:
I did a case insensitive regex grep for both `[0-9]bit` and `[0-9] bit` and only found those.

README.md
```diff
@@ -1,7 +1,22 @@
-# ETC.A
+# ETCa
 Extensible Turing Complete Architecture
 
-This repo is for designing an ISA that has a simple core and a lot of extensions that can be added ontop.
-It's primary goal is to be build in the game [Turing Complete](https://turingcomplete.game) and to be easy to implement in there.
+This repo is for designing an ISA that has a simple base and multiple extensions that can be added ontop. It's primary goal is to be easy to build in the game
```
JKTKops: Are we going for grammar fixes here?
```suggestion
This repo is for designing an ISA that has a simple base and multiple extensions that can be added on top. Its primary goal is to be easy to build in the game
```
"It's" means "it is." Confusing, but "its" is possessive and "it's" is not.

README.md
```diff
@@ -1,7 +1,22 @@
...
+[Turing Complete](https://turingcomplete.game) after the second architecture is completed. Several other goals include being educational and being practical for a real CPU.
 
-An extensible assembler built for this arch is [here](https://github.com/MegaIng/etca-asm)
+An extensible assembler built for this architecture is [here](https://github.com/ETC-A/etca-asm)
```
JKTKops:
```suggestion
An extensible assembler built for this ISA is [the etca-asm repository](https://github.com/ETC-A/etca-asm) in this organization.
```

README.md
```diff
@@ -1,7 +1,22 @@
...
+
+## Terminology
+
+_Must_: Specifies that a specific behavior is required by the specification.
+
+_Should_: Specifies that a specific behavior is recommended but not required.
+
+_May_: Specifies that a specific behavior is allowed and not encouraged or discouraged.
```
JKTKops: These are not quite right: [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)

README.md
```diff
@@ -1,7 +1,22 @@
...
+
+_Illegal_: Specifies that a specific behavior or instruction is **NOT** allowed in a compliant implementation and must be handled as specified by the ISA.
+
+_Reserved_: Specifies that the behavior or instruction will be specified in a future extension or feature and should be treated as _illegal_ until it is both specified and implemented.
```
JKTKops, line 15: Didn't we decide on something more complex here? I was under the impression that it was UB, unless the interrupts extension was available, in which case it was required to cause a `#UD` exception.

base-isa.md
```diff
@@ -1,14 +1,15 @@
 # General design
 
-- 16bit word size
-- 8 16bit registers
+- 16 bit word size
+- 16 bit address space
+- 8 16 bit registers
 - 4 flags set upon (some) ALU operations: Zero (`Z`), Negative (`N`), Carry (`C`), Overflow (`V`)
 - CPU Execution starts at address 0x8000
+  - This is only relevant if one or more extension are implemented which expose the instruction pointer in some way.
```
JKTKops:
```suggestion
  - This is not relevant for the ISA with no extensions.
```

I think the goal with this bullet point is to be easier for newcomers, right? So I think it should be as direct as possible about which situation it affects, without being so long as to be confusing.

base-isa.md
```diff
@@ -1,14 +1,15 @@
...
 
 # Base Instructions
 
-This is a group of 16bit fixed width instructions. The motivation for the chosen instructions is to be both complete and
-easy to build in game.
+This is a group of 16 bit fixed width instructions. The motivation for the chosen instructions is to be both complete and easy to build in game.
```
JKTKops:
```suggestion
This is a group of 16 bit fixed width instructions.
```
If we really want the extra sentence, it belongs in the rationale file.

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
 | First byte    | Second Byte  | Comment                                  |
 |:--------------|:-------------|:-----------------------------------------|
 | `00 01 CCCC`  | `RRR RRR 00` | 2 register computation                   |
-| `00 SS CCCC`  | `RRR RRR ??` | when `SS != 01`, reserved for extensions |
-| `00 01 CCCC`  | `RRR RRR ??` | when `?? != 00`, reserved for extensions |
-| `00 01 11??`  | `RRR RRR ??` | reserved for extensions                  |
+| `00 01 11??`  | `???? ????`  | reserved for extensions                  |
+| `00 SS CCCC`  | `RRR RRR MM` | when `MM != 00`, reserved for extensions |
```
JKTKops: GitHub won't let me annotate the whole range that I want because of the mix of +/- diff lines.

Aim to put general _legal_ patterns, followed by (possibly more than one) related patterns which are reserved, _from most general to most specific_. Overlap in those related patterns is fine; conceptually this is why most general should come first. It's easier to process. Then repeat with more legal patterns and more related reservations.

Rearrange so that we have 2 register computations, followed by the 2 register pattern with SS/MM. Specifically, from the latest commit, put line 27 after line 25. Then "when `MM != 00`" can be replaced by "when not matching above", which is more specific as it also covers the `SS` bits which are seemingly ignored here (they are mentioned below, but in a separate pattern that seems to be confusing overlap). 

Next put the current line 26, which explicitly reserves 4 opcodes so is less general than the previous pattern.

Then line 29 should remain `01 SS CCCC` (though the second byte => `???? ????` is good IMO). Similarly replace explanation with "when not matching above." The generalization of the `01` to `0?` helps compact the table but makes it harder to process.

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
...
 | `01 01 CCCC`  | `RRR IIIII`  | immediate and 1 register computation     |
-| `01 SS CCCC`  | `RRR IIIII`  | when `SS != 01`, reserved for extensions |
+| `0? SS CCCC`  | `???? ????`  | when `SS != 01`, reserved for extensions |
 | `10 0 D CCCC` | `DDDDDDDD`   | (conditional) relative jump instruction  |
 | `10 1 ?????`  | `?????????`  | reserved for extension                   |
```
JKTKops:
```suggestion
| `10 1 ?????`  | `?????????`  | reserved for extensions                  |
```

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
...
-| S      | size (res.d)                 |
-| ?      | reserved for unknown purpose |
+| Symbol | Meaning         |
+|--------|-----------------|
+| C      | Opcode          |
+| R      | Register Id     |
+| I      | Immediate       |
+| D      | Displacement    |
+| S      | Operation Size  |
+| M      | Memory Mode     |
```
JKTKops:
```suggestion
| M      | Memory Addressing Mode     |
```
I'm not so set on this one, but if we're aiming to be technically correct...

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
...
-| C      | opcode bit                   |
-| R      | register id                  |
-| I      | immediate                    |
-| D      | displacement                 |
-| S      | size (res.d)                 |
-| ?      | reserved for unknown purpose |
+| Symbol | Meaning         |
+|--------|-----------------|
+| C      | Opcode          |
+| R      | Register Id     |
```
JKTKops: `ID` or `Index` but not `Id`.

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
...
+| S      | Operation Size  |
+| M      | Memory Mode     |
+| ?      | Arbitrary Value |
 
-### Execution of Reserved Instructions
-
-Execution of a reserved instruction is undefined behavior. A future extension will change this to "the CPU must trigger a specific interrupt."
+### Execution of Illegal and Reserved Instructions
 
+Execution of an _illegal_ or _reserved_ instruction is _unspecified_ behavior. A future extension will change this to require an interrupt to occur.
```
JKTKops: This is not consistent with the text of the README. While it quite possibly makes sense in the context of other extensions being present, remember most people will read the README and then immediately read this document. If they contradict each other it would be quite confusing.

I also think lines like this should probably link to the relevant extension if it already exists. In most cases, at this point, it does.

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
...
-| ?      | reserved for unknown purpose |
+| Symbol | Meaning         |
+|--------|-----------------|
+| C      | Opcode          |
+| R      | Register Id     |
+| I      | Immediate       |
+| D      | Displacement    |
+| S      | Operation Size  |
+| M      | Memory Mode     |
+| ?      | Arbitrary Value |
```
JKTKops, line 43:
```suggestion

Notice that whenever `MM` bits are present, it is reserved for their value to be anything other than `00`. Similarly, it is reserved for `SS` bits to have values other than `01`.

```

For emphasis, since I don't think _any_ way of writing that table is going to be perfectly clear.

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
...
+Execution of an _illegal_ or _reserved_ instruction is _unspecified_ behavior. A future extension will change this to require an interrupt to occur.
 
 ## Computation Instructions
 
-Both computation formats share a lot of similarities. For both the first byte has the structure `0x SS CCCC`.
+Both computation formats share a lot of similarities. For both, the first byte has the structure `0x SS CCCC`.
 
-- `SS` is a size marker and reserved for extensions. For the base instructions this should always be `01`
-- `CCCC` is a 4bit opcode deciding which operation to execute. This can be an ALU or a memory load/store instruction
+- `SS` is a size marker and reserved for extensions. For the base instructions this _must_ always be `01`
```
JKTKops:
```suggestion
- `SS` is a size marker and reserved for extensions. For the base instructions this is always `01`.
```

This isn't a specification of instruction semantics, but rather of what instructions exist in the first place.

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
...
+- `CCCC` is a 4 bit opcode deciding which operation to execute.
 
 ### 2 Register Computation
 
-The second byte has the format `AAA BBB 00` where `AAA` and `BBB` are references to registers. `AAA` is the destination register and left source register. `BBB` is the right source register. The only exception is the RSUB instruction where `AAA` and `BBB` are swapped for the purpose of being source registers.
+The second byte has the format `AAA BBB MM` where `AAA` and `BBB` are references to registers. `AAA` is the destination register and left source register. `BBB` is the right source register. There are 2 exceptions to this. The first is the RSUB instruction where `AAA` and `BBB` are swapped for the purpose of being source registers. The second exception is that the `B` argument specifies which memory address or control register to write to. `MM` is the memory mode which will be defined in an extension and is only defined for `00` in the base isa.
```
JKTKops:
```suggestion
The second byte has the format `AAA BBB 00` where `AAA` and `BBB` are references to registers. `AAA` is the destination register and left source register. `BBB` is the right source register. The only exception is the RSUB instruction where `AAA` and `BBB` are swapped for the purpose of being source registers.
```

(Suggesting revert) No need to complicate the specification of base by explaining how it interacts with an extension that it doesn't require...

References to "memory" only apply to non-base instructions (`MM != 00`) which aren't even defined here. For all a reader knows, those formats could be _completely_ different (and, in a sense, they are, since some data is shifted to a third byte in many cases). `BBB` never specifies a control register in this format, because `READCR` and `WRITECR` opcodes do not have 2r forms.

base-isa.md
```diff
@@ -22,48 +23,45 @@ The highest two bits of the first byte are a format marker:
...
 
 ### Immediate Computation
 
-The second byte has the format `RRR IIIII`, where the 5bit immediate acts as operand `B`. The immediate is sign extended for operations 0-7 and operation 9. The immediate is zero extended for operation 8 and operations 10-15. While interleaving of signed and unsigned immediate at the border of the 2 sections seems unusual at first glance, it's done to make decoding hardware simpler.
+The second byte has the format `RRR IIIII`, where the 5 bit immediate acts as operand `B`. The immediate is sign extended for operations 0-7 and operation 9. The immediate is zero extended for operation 8 and operations 10-15.
```
JKTKops: Since we've recently established that operations 8 and 9 are equally easy to detect, I'm completely in favor of swapping `movs` and `movz` back to the way they used to be. This is a breaking change but simplifies documentation.

Also `movs` is probably more common in practice so my brain likes the idea of it having opcode `8` because that number is "nicer." My brain is weird, and I have no idea how much this is biasing my favor for swapping those opcodes back :P

base-isa.md
```diff
@@ -85,17 +83,17 @@ TODO: This is a baseline, very much still floating
...
-4) Primary intent is that these are used with immediate. Exact assignment of control registers is still floating. At least the the CPU status/extension/feature control registers should be present.
-5) The C and V flags are in an undefined state after execution of these instructions. Implementations may do whatever is easiest. An extension may mandate a particular behavior, with good enough reason, but must *not* mandate that the value of these flags after the operation depends on their value before the operation.
-6) These instructions do not have a 2 register mode. The corresponding bit patterns (`00 SS 11??`) for the first byte are reserved. This can easily be detected by using similar to 2)
-7) This instruction zero extends argument B as if the value read is of the size specified by the SS bits in the instruction (assuming a relevant extension is present)
-8) This instruction sign extends argument B as if the value read is of the size specified by the SS bits in the instruction (assuming a relevant extension is present)
+4) Control registers can only be accessed with immediates as a way to prevent potential security vulnerabilities.
```
JKTKops:
```suggestion
4) Control registers can only be accessed with immediates.
```

Rationale belongs in the rationale document.

base-isa.md
```diff
@@ -85,17 +83,17 @@ TODO: This is a baseline, very much still floating
...
-8) This instruction sign extends argument B as if the value read is of the size specified by the SS bits in the instruction (assuming a relevant extension is present)
+4) Control registers can only be accessed with immediates as a way to prevent potential security vulnerabilities.                                                                  
+5) The C and V flags are in an _unspecified_ state after execution of these instructions. Extensions _may_ mandate a particular behavior, with good enough reason, but must **NOT** mandate that the value of these flags after the operation depends on their value before the operation.
+6) These instructions do not have a 2 register mode. The corresponding bit patterns (`00 SS 11??`) for the first byte are _reserved_. This can easily be detected by using a similar method to 2).
+7) This instruction zero extends argument B as if the value read is of the size specified by the SS bits in the instruction (assuming a relevant extension is present).
+8) This instruction sign extends argument B as if the value read is of the size specified by the SS bits in the instruction (assuming a relevant extension is present).
```
JKTKops: I think these can probably be combined into "These instructions are equivalent in the base ISA. Extensions which allow `SS` values other than `01` make these instructions distinguishable."

This simplifies the explanation significantly for someone who is only reading `base` while also very briefly justifying having both of them without explaining the nitty gritty of why (which requires some familiarity with the relevant extensions in order to make sense). Links to the 3 relevant extensions would also make sense.

base-isa.md
```diff
@@ -108,26 +106,24 @@ Undefined control registers are reserved and reading from or writing to them is
 
 ## Memory Semantics
 
-Unaligned memory accesses are undefined behavior. A memory access is unaligned if the address modulus the read/write width is not equal to zero. The read/write width is defined by the SS bits, which for the base ISA corresponds to a read/write width of 2 bytes.
+Unaligned memory accesses are _unspecified_ behavior. A memory access is unaligned if the address modulus the read/write width is not equal to zero. The read/write width is defined by the SS bits, which for the base ISA corresponds to a read/write width of 2 bytes.
```
JKTKops: This is another sentence that I think benefits from restricting context to base-isa. "A memory access is unaligned if the address is not divisible by 2." The `SS` extensions could be modified to mention how they affect alignment requirements if they don't already.

base-isa.md
```diff
@@ -108,26 +106,24 @@ Undefined control registers are reserved and reading from or writing to them is
...
 
 All IO is memory mapped
 
 ### Separation of Program ROM and RAM
 
-Aside from program rom starting at address 0x8000, there are no requirements for how the memory address space is layed out. For ease of implementation, this spec does NOT require program memory to be accessible with the load and store instructions nor does it require that RAM be executable. Loads and stores of program memory are undefined behavior. Executing from RAM is undefined behavior. Future extensions will change this behavior
+Aside from program rom starting at address 0x8000, there are no requirements for how the memory address space is layed out. For ease of implementation, this spec does **NOT** require program memory to be accessible with the load and store instructions nor does it require that RAM be executable. Loads and stores of program memory are _unspecified_ behavior. Executing from RAM is _unspecified_ behavior. Future extensions and/or features will change this behavior
```
JKTKops:
```suggestion
Aside from program rom starting at address 0x8000, there are no requirements for how the memory address space is layed out. For ease of implementation, this spec does **NOT** require program memory to be accessible with the load and store instructions nor does it require that RAM be executable. Loads and stores of program memory are _unspecified_ behavior. Executing from RAM is _unspecified_ behavior. Future extensions and/or features will change this behavior.
```

base-isa.md
```diff
@@ -108,26 +106,24 @@ Undefined control registers are reserved and reading from or writing to them is
...
 
 ## Flag Semantics
 
-Flags conceptually maintain these values. A future, more rigorous specification of this standard will include a detailed description of affected flags for every instruction individually.
-
 1) The `Z` flag indicates that the result was zero.
 2) The `N` flag indicates that the result, interpreted as a 2's complement number, was negative.
 3) The `C` flag describes whether an additive instruction caused a carry, or a subtractive computation caused a borrow. The concept of "borrow" is exactly the same as in long-hand decimal subtraction the way you might do it on paper. The flag is set if the subtraction would borrow out of the next most significant bit. A borrow happens precisely when the corresponding addition does _not_ carry. In other words, in order to subtract, you (a) complement the B input, (b) set the carry in, (c) set C to the inverse of the carryout.
```
JKTKops:
```suggestion
3) The `C` flag describes whether an additive instruction caused a carry, or a subtractive computation caused a borrow. The concept of "borrow" is exactly the same as in long-hand decimal subtraction the way you might do it on paper. The flag is set if the subtraction would borrow out of the next most significant bit. A borrow happens precisely when the corresponding 2's-complement addition does _not_ carry. In other words, in order to subtract, you (a) complement the B input, (b) set the carry-in, (c) set C to the inverse of the carryout.
```

base-isa.md
```diff
@@ -108,26 +106,24 @@ Undefined control registers are reserved and reading from or writing to them is
...
-4) The `V` flag indicates that the operation caused overflow. This means that storing the result in the destination register caused a loss of precision.
+4) The `V` flag indicates that the operation caused overflow. This means that a loss of precision occurred for a signed integer calculation (eg. positive + positive = negative or negative + negative = positive).
 
 ## Jump Instructions
 
-Here the first byte has the format `10 0 D CCCC` where `D` fills the high byte of the displacement. `CCCC` is the condition to check. The second byte is a single 8bit immediate representing the low byte of the displacement. This combined 9 bit displacement (sign extended to 16 bits) is added to the base address of the current instruction and stored in the program counter.
+Here the first byte has the format `10 0 D CCCC` where `D` fills the high byte of the displacement. `CCCC` is the condition to check. The second byte is a single 8 bit immediate representing the low byte of the displacement. This combined 9 bit displacement (sign extended to the address width which is 16 bits in the base isa) is added to the base address of the current instruction and stored in the program counter.
```
JKTKops:
```suggestion
Here the first byte has the format `10 0 D CCCC` where `D` fills the high byte of the displacement. `CCCC` is the condition to check. The second byte is a single 8 bit immediate representing the low byte of the displacement. This combined 9 bit displacement (sign extended to the address width, which is 16 bits in the base isa) is added to the base address of the current instruction and stored in the program counter.
```

extensions/cache-instructions/README.md
```diff
@@ -24,9 +24,13 @@ are present.
 
 Note that the last 5 instructions overlap with the never jump instruction. This is intentional since these instructions don't change a programs behavior.
 
+If this extension is present on a system without a data cache, `ALLOC_ZERO` must write 0 to memory as if it had a cache line size as specified by the `CACHE_LINE_SIZE` control register.
+If `CACHE_LINE_SIZE` is zero, then it _may_ be a NOP instruction.
```
JKTKops:
```suggestion
If `CACHE_LINE_SIZE` is zero, then it _must_ be a NOP instruction.
```
Surely... otherwise code has to special case this.

extensions/full-immediates/README.md
```diff
@@ -44,11 +44,11 @@ the following table:
...
 If the [register expansion](../expanded-registers/README.md) extension is available, then `REX.Q` can be used with instructions from this
 extension. It has an effect only if both of the following conditions are met:
-  - The instruction is an `iS` mode
-  - The instruction's Operand Size attribute is `quad`
+  - The instruction has an `iS` operand
+  - The instruction's operand size attribute is `quad`
 Then the immediate is an 8-byte literal value instead of a 4-byte literal value.
 
-Sign extension of the immediate follows the same rules as used for 5 bit immediates
+Sign extension of the immediate follows the same rules as used for 5 bit immediate in register-immediate mode.
```
JKTKops:
```suggestion
Sign extension of the immediate follows the same rules as used for 5 bit immediates in register-immediate mode.
```

extensions/interrupts/README.md
```diff
@@ -45,7 +45,7 @@ The following table specifies which flag is associated with which bit in the `FL
...
 | System Call              | Synchronous  | 0                | Used by programs to ask the operating system to do something.                                                      |
 | Timer                    | Asynchronous | 1                | Occurs when a timer causes an interrupt.                                                                           |
-| Illegal Instruction      | Synchronous  | 2                | Occurs when execution of an undefined, reserved, or illegal instruction is attempted.                              |
+| Illegal Instruction      | Synchronous  | 2                | Occurs when execution of a reserved, or illegal instruction is attempted.                                          |
 | Memory Alignment Error   | Synchronous  | 3                | Occurs when an attempt to read or write to memory at an unaligned address occurs and is not supported.             |
 | General Protection Fault | Synchronous  | 4                | Occurs when the CPU needs the operating system to handle an unexpected event for stability or consistency reasons. |
```
JKTKops, line 50: Do we care about specifying/explaining the difference between an _interrupt_ (something needs to happen, but the program is fine) and an _exception/fault_ (the program has done something wrong and needs to be corrected or terminated)?

What about giving these convenient (but suggestive) abbreviations for reference elsewhere, such as `#UD`, `#AC`, and `#GP`? Syntax ala x86, but just a placeholder.

extensions/quad-word-operations/README.md
```diff
@@ -15,4 +15,4 @@
 
 # Assembly changes
 
-64bit register references/64bit operations are marked by the infix/prefix `q` (i.e. `%rq0`)
+64 bit register references/64 bit operations are marked by the infix/prefix `q` (i.e. `%rq0`)
```
JKTKops:
```suggestion
64 bit register references/64 bit operations are marked by the infix/prefix `q` (i.e. `%rq0`).
```

features/README.md
```diff
@@ -7,5 +7,4 @@
 |     2     | [Cache Coherency](./cache-coherency)                      | Under Development |
 
 
-Features are a second type of extension that __ONLY__ change behavior of the CPU without adding instructions, control registers, or any other similar aspect. They generally mandate a specific behavior when certain actions occur but do __NOT__
-change how instructions are decoded.
+Features are a second type of extension that __ONLY__ change behavior of the CPU without adding instructions, control registers, or any other similar aspect. They generally only mandate a specific behavior when certain actions occur.
```
JKTKops:
```suggestion
Features are a second type of extension that __ONLY__ change behavior of the CPU without adding instructions, control registers, or any other similar aspect. They generally only mandate a specific behavior when certain otherwise unspecified actions occur.
```
clarity?

JKTKops (changes requested):

README.md
EnderShadow: In the base ISA, both illegal and reserved are undefined behavior. The interrupt extension mandates that illegal (and by extension reserved) causes an interrupt.

base-isa.md
EnderShadow: How is this inconsistent?

base-isa.md
EnderShadow: Mega had suggested that we explicitly mention how the `BBB` acts as a memory address/control register ID since others had gotten confused about it.

base-isa.md
EnderShadow: I'm against changing it because it doesn't affect decoding in any significant way AND is a breaking change.

extensions/interrupts/README.md
EnderShadow: I think differentiating between an interrupt and exception is probably a good idea. I'm neutral on creating abbreviations.

base-isa.md
MegaIng: `STORE` also uses ``BBB`` as the destionation and does have a 2 register form. This entire section is actually also not only applicable to the 2-reg addressing, but also for the reg-imm addressing. Maybe it shouldn't be called `BBB` and `AAA` here then.

base-isa.md
MegaIng: If we decide to change how `test` works, we can IMO also at the same time swap movs and movz.

README.md
MegaIng: These sound like simplifications, but not like they are fundamentally wrong (except maybe `May`). Maybe just saying these are summaries and linking the RFC is a good idea.

extensions/interrupts/README.md
MegaIng: I agree with Ender in both points.

README.md
EnderShadow: I'll be honest, I didn't know where the standard for this was defined. I just went off memory from what I've seen. I agree that `May` should be fixed and the RFC should be linked.

base-isa.md
EnderShadow: I remembered my reasoning for having movz at 8 and movs at 9. It's because you can use the 1s bit directly in an AND gate without needing to negate it first. So it saves a single gate.

JKTKops:
Can we get an update on this? Might as well finish it up and merge. I think we should aim to have specification issues closed by the time custom levels release in TC.

EnderShadow:
I will as soon as I take a look at it again which will be at least a week since I'm out of town

base-isa.md
EnderShadow: Done, please take a look

JKTKops: LGTM

base-isa.md
JKTKops: I don't think breaking is significant since it's indistinguishable in base and there is 1 single nontrivial ETCa program out there written for a relevant extension. That said, I agree with Mega; it can stay for now.

base-isa.md
JKTKops: This section is right under the "2 register computation" header. Are you proposing putting it somewhere else?

base-isa.md
```diff
@@ -19,50 +20,62 @@ The highest two bits of the first byte are a format marker:
...
+The second byte has the format `AAA IIIII`, where the 5 bit immediate acts as operand `B`. The immediate is sign extended for operations 0-7 and operation 9. The immediate is zero extended for operation 8 and operations 10-15. `AAA` is treated the same as in the 2 Register Computation section above.
 
-### Opcodes
+### Exceptions
 
-TODO: This is a baseline, very much still floating
+There are several exceptions to how the operands work.
+
+ - `RSUB`
+    - `BBB` is the left source register and `AAA` is the right source register.
```
MegaIng: Should it be mentioned that this also applies to Immediate mode? (So `IIIII` or just `B` instead of `BBB`?) or maybe it's ok like this, IDK.

EnderShadow: Good catch!

MegaIng (approved):

base-isa.md
EnderShadow: I moved the exceptions to there own section. How's it look?

JKTKops (approved):
LGTM, very exciting :D

PR #41, opened 2023-01-05
Initial spec of real 32-bit address mode

### LINKED ISSUE #18: Memory operands 1 extension references the EXTEN control register which is no longer present
https://github.com/ETC-A/etca-spec/tree/main/extensions/memory-operands-1#d8-and-dp

This will need to be fixed when we draft the larger address space extensions

JKTKops:
fixes #18

Initial specification, probably best to merge this but continue discussion?

We should probably generally have a clearer system for discussion. Perhaps an issue for each extension which is not finalized, and a page under projects with direct links to each of them.

### DIFF
--- a/extensions/32-bit-address-space/README.md
+++ b/extensions/32-bit-address-space/README.md
@@ -12,10 +12,6 @@ processor is actually in will almost certainly behave incorrectly.
 
 In particular, this extension focuses on the _Real 32-bit address mode_.
 
-# Added Instructions
-
-None!
-
 # Modes
 
 CR `1110`, equivalently `cr14` or "the MODE control register," holds a value indicating the current
--- a/extensions/32-bit-address-space/README.md
+++ b/extensions/32-bit-address-space/README.md
@@ -1,8 +1,8 @@
 # General Design
 
-**Extension State: Under Development**
-**Requires: CP1.14**
-**CP Bit: CP1.16**
+**Extension State: Under Development**  
+**Requires: CP1.14**  
+CPUID Bit: CP1.16**
 
 # Overview
 
--- a/extensions/32-bit-address-space/README.md
+++ b/extensions/32-bit-address-space/README.md
@@ -31,7 +31,7 @@ In real 32-bit address mode, all addresses are treated as being 32 bits long. En
 address mode _must_ preserve the program counter. Returning to Bare mode from real 32-bit address
 mode must preserve the program counter _if possible_ - if the current program counter interpreted
 as a 16-bit address would not be the address of the current instruction, the behavior of the
-system is undefined.
+system is _unspecified_.
 
 # Interactions With Other Extensions
 
--- a/extensions/32-bit-address-space/README.md
+++ b/extensions/32-bit-address-space/README.md
@@ -1,0 +1,58 @@
+# General Design
+
+**Extension State: Under Development**
+**Requires: CP1.14**
+**CP Bit: CP1.16**
+
+# Overview
+
+This extension adds a new concept of _modes_ to the ISA. Modes are used to control things about the processor
+state which are binary-incompatible - that is, binary code written for a mode other than the mode the
+processor is actually in will almost certainly behave incorrectly.
+
+In particular, this extension focuses on the _Real 32-bit address mode_.
+
+# Added Instructions
+
+None!
+
+# Modes
+
+CR `1110`, equivalently `cr14` or "the MODE control register," holds a value indicating the current
+_mode_ of the processor.
+
+## Bare Mode
+
+The mode described by [the base isa](../../base-isa.md) is known as the
+Bare mode (Real mode?). In that mode, pointers are 16 bits and there is no virtual memory, paging,
+memory protection, safe multitasking, or privilege restriction. Bare mode is indicated by a value of
+0 in `cr14`.
+
+## Real 32-bit Address Mode
+
+A new mode known as _Real 32-bit address mode_ is added, indicated by a value of 2 in `cr14`.
+In real 32-bit address mode, all addresses are treated as being 32 bits long. Entering real 32-bit
+address mode _must_ preserve the program counter. Returning to Bare mode from real 32-bit address
+mode must preserve the program counter _if possible_ - if the current program counter interpreted
+as a 16-bit address would not be the address of the current instruction, the behavior of the
+system is undefined.
+
+# Interactions With Other Extensions
+
+Other extensions mention "address modes," in particular when deciding what a particular value
+should mean or how many words an immediate should be. The address mode of Real 32-bit Address Mode
+is `doubleword`.
+
+(The above language is chosen because other values of the MODE register may also specify
+a `doubleword` address mode)
+
+# Recommendations
+
+This extension does not _require_ behaviors beyond that. However, for compatibility with future
+extensions which will require more specifics, the following are recommended:
+
+* Program execution should start at (real) address `0xFFFF8000`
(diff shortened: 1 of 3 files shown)


extensions/32-bit-address-space/README.md
```diff
@@ -0,0 +1,58 @@
+# General Design
+
+**Extension State: Under Development**
+**Requires: CP1.14**
+**CP Bit: CP1.16**
```
EnderShadow: Add trailing double space for proper formatting
```suggestion
**Extension State: Under Development**  
**Requires: CP1.14**  
CPUID Bit: CP1.16**
```

extensions/32-bit-address-space/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Overview
+
+This extension adds a new concept of _modes_ to the ISA. Modes are used to control things about the processor
+state which are binary-incompatible - that is, binary code written for a mode other than the mode the
+processor is actually in will almost certainly behave incorrectly.
+
+In particular, this extension focuses on the _Real 32-bit address mode_.
```
EnderShadow, line 13: Do we need to mention "Real" here since there's no other defined 32 bit addressing mode yet?

extensions/32-bit-address-space/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Added Instructions
+
+None!
+
+# Modes
```
EnderShadow: No need to mention added instructions if there are none.
```suggestion
# Modes
```

extensions/32-bit-address-space/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+CR `1110`, equivalently `cr14` or "the MODE control register," holds a value indicating the current
+_mode_ of the processor.
+
+## Bare Mode
+
+The mode described by [the base isa](../../base-isa.md) is known as the
+Bare mode (Real mode?). In that mode, pointers are 16 bits and there is no virtual memory, paging,
+memory protection, safe multitasking, or privilege restriction. Bare mode is indicated by a value of
+0 in `cr14`.
```
EnderShadow: Maybe this should be called base mode or initial mode? Also, privilege restriction has it's own CR so privilege restrictions can still occur without this. Does safe multitasking need to be explicitly mentioned? Isn't that a combined property of interrupts + memory protection?

extensions/32-bit-address-space/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Real 32-bit Address Mode
+
+A new mode known as _Real 32-bit address mode_ is added, indicated by a value of 2 in `cr14`.
+In real 32-bit address mode, all addresses are treated as being 32 bits long. Entering real 32-bit
+address mode _must_ preserve the program counter. Returning to Bare mode from real 32-bit address
+mode must preserve the program counter _if possible_ - if the current program counter interpreted
+as a 16-bit address would not be the address of the current instruction, the behavior of the
+system is undefined.
```
EnderShadow: Out of curiosity, why the value 2 instead of the value 1? 

```suggestion
A new mode known as _Real 32-bit address mode_ is added, indicated by a value of 2 in `cr14`.
In real 32-bit address mode, all addresses are treated as being 32 bits long. Entering real 32-bit
address mode _must_ preserve the program counter. Returning to Bare mode from real 32-bit address
mode must preserve the program counter _if possible_ - if the current program counter interpreted
as a 16-bit address would not be the address of the current instruction, the behavior of the
system is _unspecified_.
```

extensions/32-bit-address-space/README.md
```diff
@@ -0,0 +1,58 @@
...
+# Recommendations
+
+This extension does not _require_ behaviors beyond that. However, for compatibility with future
+extensions which will require more specifics, the following are recommended:
+
+* Program execution should start at (real) address `0xFFFF8000`
+* Addresses in Bare mode refer to the (real) address which is their sign extension to 32 bits
+* The program counter (or equivalent) in the processor itself should always store the sign-
+  extended address when operating in Bare mode. This way, nothing needs to be done when
+  entering or leaving real 32-bit address mode.
```
EnderShadow: Aren't those first 2 points required by the spec?
```suggestion
* Program execution should start at (real) address `0xFFFF8000`
* Addresses in Bare mode refer to the (real) address which is their sign extension to 32 bits
* The program counter (or equivalent) in the processor itself should always store the sign-
  extended address when operating in Bare mode. This way, nothing needs to be done when
  entering or leaving real 32-bit address mode from or to Bare mode.
```

EnderShadow (changes requested):
You also need to update README for the extensions list

extensions/32-bit-address-space/README.md
```diff
@@ -0,0 +1,58 @@
...
+In particular, this extension focuses on the _Real 32-bit address mode_.
+
+# Added Instructions
+
+None!
+
+# Modes
+
+CR `1110`, equivalently `cr14` or "the MODE control register," holds a value indicating the current
+_mode_ of the processor.
```
EnderShadow, line 18: This should also state that it's only accessible in system privilege mode

extensions/32-bit-address-space/README.md
JKTKops: I think we should, because eventually there _will_ be a protected (or virtual, whatever we decide to call it) 32-bit addressing mode. But I'm open to names other than "real."

extensions/32-bit-address-space/README.md
JKTKops: "Base" mode seems like a good name. I'll update that.
I think in terms of "safe multitasking" I was thinking more like multicore cache ownership and stuff like that. But yeah, as long as the system supports caching in real mode at all, it seems orthogonal to this, so I'll remove that bit. Simply no virtual memory, paging, or memory protection (as memory protection is typically part of paging).

extensions/32-bit-address-space/README.md
JKTKops: As per discussion on Discord, bit 2 was used so that bit 1 could stay free to indicate virtual memory. I think this point is settled now?

extensions/32-bit-address-space/README.md
JKTKops: No. The second point is _certainly_ not, because base doesn't talk about 32 bit addresses. The first point is therefore not either.

The first point is required by the combination of base and the second point, but recall that the starting address is unobservable in base, so I think it is OK to reiterate it. Perhaps I should mention that it is simply being reiterated and would be required anyway, i.e. nothing special needs to happen?

extensions/32-bit-address-space/README.md
JKTKops: If privileges are supported, sure.

extensions/32-bit-address-space/README.md
EnderShadow: Maybe we should define what "real" means in this context for anyone who isn't familiar with the terminology?

JKTKops: I think not. In a vacuum it's just a name that doesn't have to mean anything. It's only meaningful when contrasted against virtual or segmented memory.

I tried describing it, realized that that essentially requires saying "in contrast to virtual memory" (because why would an address _not_ refer to that location in physical memory?) and then gave up. I think the virtual memory extension is the place to point out that something clever can happen.

MegaIng (approved):

EnderShadow (approved):

extensions/32-bit-address-space/README.md
EnderShadow: That's a fair point

PR #46, opened 2023-03-06
Minor fixes

EnderShadow:
I moved halt out of the non-canonical NOP section since HALT does not behave as a NOP for a program which is unaware of the HALT instruction.


MegaIng (approved):

EnderShadow:
Thanks @MegaIng

PR #47, opened 2023-03-12
Add multiple memory access instructions feature

EnderShadow:


MegaIng (dismissed):

features/multiple-memory-access-instructions/README.md
```diff
@@ -0,0 +1,11 @@
+# General design
+
+**Feature State: Mostly Stable**  
+**FEAT Bit: 3**
+
+* The memory instructions, including but not limited to `store`, `load`, `push`, and `pop`, _must_ accept memory operands as specified by the `MO1` and `MO2` extensions.
+* The `pop` instruction _must_ be treated as an _illegal_ instruction when the RMI byte follows the pattern `xxx 00x 01`.
```
JKTKops, line 8: Please specify what this is preventing for clarity.

extensions/memory-operands-2/README.md
```diff
@@ -49,6 +49,10 @@ The behavior when `REX.Q` is set while in `quad` addressing mode when _any_ of t
   - The instruction has an `iS` and a `d8`
   - The instruction has a `dP` and an `i8`
 
+## Exceptions
+
+Using this extension with any instruction that would normally access memory, including but not limited to `store`, `load`, `push`, and `pop`, is _unspecified_ behavior.
```
JKTKops: This is specified (with less clear wording IMO) right underneath this. Can we pick one or the other and not both?

extensions/memory-operands-2/README.md
```diff
@@ -49,6 +49,10 @@ The behavior when `REX.Q` is set while in `quad` addressing mode when _any_ of t
   - The instruction has an `iS` and a `d8`
   - The instruction has a `dP` and an `i8`
 
+## Exceptions
```
JKTKops: I'm generally hesitant to overload the term "Exceptions" since hardware/ISA exceptions are different from natural-language exceptions. These instructions don't cause exceptions, discounting memory access faults inherent to the memory operand, which isn't what is discussed in this section. Maybe something like "Unaffected Instructions"?

JKTKops (changes requested):

features/multiple-memory-access-instructions/README.md
EnderShadow: Does the latest commit explain it well enough?

extensions/memory-operands-2/README.md
EnderShadow: Looks like only MO2 specifies it twice. Fixed

extensions/memory-operands-2/README.md
EnderShadow: Understandable. I removed the exceptions section

extensions/memory-operands-1/README.md
```diff
@@ -99,6 +99,9 @@ as though the operand values were the size specified by the instruction's Operan
 
 # Added Instruction
 
+These instruction formats apply to any instruction which encodes operands using an `ABM` byte _unless_ that instruction already performs a memory
```
MegaIng: This is still duplicated, it is already mentioned above at the start of "Added instruction formats". MO1 and MO2 are inconsistent in that regard.

features/multiple-memory-access-instructions/README.md
MegaIng: Does this means it's legal to use a memory location as a stack pointer? That would mean three memory accesses, right?

EnderShadow: Yes it would. I think we should probably specify that it would only be usable as a stack pointer if they also have the arbitrary stack pointer extension?

MegaIng: Even then it's kinda weird IMO. But yeah, asp should probably be a requirement for this.

EnderShadow: I added it in the feature

extensions/memory-operands-1/README.md
EnderShadow: 🤦
Should be fixed now...

extensions/memory-operands-2/README.md
EnderShadow: I stand corrected. They both had it defined twice because of an inconsistency in where it was located in the docs.

MegaIng (approved):

JKTKops (approved):

PR #50, opened 2023-03-24
Fixes and additions

EnderShadow:
- Change the addressing mode control register to 17 because 14 was already taken by the cache line size control register and the 2 added cache line control registers should be contiguous with the first.
- Add 2 control registers to the cache instruction extension to allow for specifying a region of memory which should not be cached. Useful for system to mark where IO is located in memory.
- Add a bypass cache prefix to the cache instructions extension since that may be required by a system ROM to properly initialize. I'm not sure if this is actually needed since since the `NO_CACHE` control registers default to setting the first cache line as non-cacheable memory.

### DIFF
--- a/extensions/cache-instructions/README.md
+++ b/extensions/cache-instructions/README.md
@@ -39,11 +39,14 @@ If `CACHE_LINE_SIZE` is zero, then it _must_ be a NOP instruction.
 in which case it _may_ be 0.
 
 `NO_CACHE_START` is a cache-aligned address which represents the inclusive start of a contiguous range of physical memory addresses which will not be cached when accessed. If the privilege extension is present, this is only accessible in system mode.
-This control register is set to 0 on CPU initialization. When writing to this control register, the value will be sign extended from the current physical address width to the maxiumum supported physical address width. Writing a non-cache-aligned value
-is _unspecified_ behavior.
+This control register is set to 0 on CPU initialization.
 
 `NO_CACHE_END` is a cache-aligned address which represents the inclusive end of a contiguous range of physical memory addresses which will not be cached when accessed. If the privilege extension is present, this is only accessible in system mode.
-This control register is set to -`CACHE_LINE_SIZE` on CPU initialization. When writing to this control register, the value will be sign extended from the current physical address width to the maximum supported physical address width. Writing a
-non-cache-aligned value is _unspecified_ behavior.
+This control register is set to `-CACHE_LINE_SIZE` on CPU initialization.
 
-If `NO_CACHE_START` is larger than `NO_CACHE_END`, the non-cacheable address range wraps past the end of the address space back to the beginning.
+## Notes on `NO_CACHE_START` and `NO_CACHE_END`
+
+- The purpose of the initial values for these control registers is to intially disable caching since the location of MMIO is unknown.
+- When writing to this control register, the value will be sign extended from the current physical address width to the maxiumum supported physical address width.
+- Writing a non-cache-aligned value is _unspecified_ behavior
+- If `NO_CACHE_START` is larger than `NO_CACHE_END`, the non-cacheable address range wraps past the end of the address space back to the beginning.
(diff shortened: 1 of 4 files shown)


extensions/cache-instructions/README.md
```diff
@@ -27,11 +27,27 @@ Note that the last 5 instructions overlap with the never jump instruction. This
...
+| `0 1111` | `NO_CACHE_START`  |
+| `1 0000` | `NO_CACHE_END`    |
 
 `CACHE_LINE_SIZE` is a read-only control register which specifies the number of bytes in a cache line for the data cache. It _must_ be a power of 2 unless no data cache is present
 in which case it _may_ be 0.
+
+`NO_CACHE_START` is a cache-aligned address which represents the inclusive start of a contiguous range of memory addresses which will not be cached when accessed. If the privilege extension is present, this is only accessible in system mode.
+This control register is set to 0 on CPU initialization.
+
+`NO_CACHE_END` is a cache-aligned address which represents the exclusive end of a contiguous range of memory addresses which will not be cached when accessed. If the privilege extension is present, this is only accessible in system mode.
```
MegaIng: This means the last address can never be marked as no cache, and it's also impossible to signal than nothing should ever be cached.

EnderShadow: No, you can set all memory as cached by setting them to the same value since the END is exclusive. We could make it inclusive to allow all memory to be marked as non-cacheable, but that seems like something that almost certainly would never occur.

extensions/cache-instructions/README.md
```diff
@@ -29,9 +29,21 @@ If `CACHE_LINE_SIZE` is zero, then it _must_ be a NOP instruction.
...
 in which case it _may_ be 0.
+
+`NO_CACHE_START` is a cache-aligned address which represents the inclusive start of a contiguous range of physical memory addresses which will not be cached when accessed. If the privilege extension is present, this is only accessible in system mode.
+This control register is set to 0 on CPU initialization. When writing to this control register, the value will be sign extended from the current physical address width to the maxiumum supported physical address width. Writing a non-cache-aligned value
+is _unspecified_ behavior.
+
+`NO_CACHE_END` is a cache-aligned address which represents the inclusive end of a contiguous range of physical memory addresses which will not be cached when accessed. If the privilege extension is present, this is only accessible in system mode.
+This control register is set to -`CACHE_LINE_SIZE` on CPU initialization. When writing to this control register, the value will be sign extended from the current physical address width to the maximum supported physical address width. Writing a
```
JKTKops: I had to read this twice before I noticed that it means "the negative of `CACHE_LINE_SIZE`." Maybe the `-` should be in the code block? Additionally, or maybe just alternatively, a sentence explaining that the effect of this is to initially disable the cache would be good.

JKTKops (dismissed):

MegaIng (dismissed):

extensions/cache-instructions/README.md
```diff
@@ -39,11 +39,14 @@ If `CACHE_LINE_SIZE` is zero, then it _must_ be a NOP instruction.
...
 `NO_CACHE_END` is a cache-aligned address which represents the inclusive end of a contiguous range of physical memory addresses which will not be cached when accessed. If the privilege extension is present, this is only accessible in system mode.
-This control register is set to -`CACHE_LINE_SIZE` on CPU initialization. When writing to this control register, the value will be sign extended from the current physical address width to the maximum supported physical address width. Writing a
-non-cache-aligned value is _unspecified_ behavior.
+This control register is set to `-CACHE_LINE_SIZE` on CPU initialization.
 
-If `NO_CACHE_START` is larger than `NO_CACHE_END`, the non-cacheable address range wraps past the end of the address space back to the beginning.
+## Notes on `NO_CACHE_START` and `NO_CACHE_END`
+
+- The purpose of the initial values for these control registers is to intially disable caching since the location of MMIO is unknown.
+- When writing to this control register, the value will be sign extended from the current physical address width to the maxiumum supported physical address width.
```
JKTKops:
```suggestion
- The purpose of the initial values for these control registers is to intially disable caching since the location of MMIO is unknown.
- When writing to these control registers, the value will be sign extended from the write width to the maxiumum supported physical address width.
```

current address width => write width is unlikely to matter if people are doing reasonable things and should simplify implementation quite a bit.

JKTKops (changes requested):

JKTKops (approved):

PR #51, opened 2023-03-28
Specify reset semantics

EnderShadow:


extensions/32-bit-address-space/README.md
```diff
@@ -15,7 +15,7 @@ In particular, this extension focuses on the _Real 32-bit address mode_.
 # Modes
 
 CR `1 0001`, equivalently `cr17` or "the MODE control register," holds a value indicating the current
-_mode_ of the processor.
+_mode_ of the processor. This has an initial value of `0`
```
JKTKops:
```suggestion
_mode_ of the processor. This has an initial value of `0`.
```

JKTKops (dismissed):
We can easily adjust these later if there's a better suggestion so pretty much blanket approval. Just one missing period I want to catch since otherwise it would be a whole 'nother PR to fix 🙃

MegaIng (approved):

JKTKops (approved):

PR #53, opened 2023-04-28
Swap order of inc/dec in push/pop instructions

JKTKops:
https://discord.com/channels/828292123936948244/946806826756882552/1101428607802421318

feel free to try and clarify the language... pre/post inc/dec ordering is always so confusing.


EnderShadow (approved):
Looks good. I don't see any issues with it.

PR #55, opened 2023-04-29
Add Real64 extension

EnderShadow:


extensions/64-bit-address-space/README.md
```diff
@@ -0,0 +1,58 @@
...
+This extension adds a new concept of _modes_ to the ISA. Modes are used to control things about the processor
+state which are binary-incompatible - that is, binary code written for a mode other than the mode the
+processor is actually in will almost certainly behave incorrectly.
+
+In particular, this extension focuses on the _Real 64-bit address mode_.
+
+# Modes
+
+CR `1 0001`, equivalently `cr17` or "the MODE control register," holds a value indicating the current
+_mode_ of the processor. This has an initial value of `0`.
```
JKTKops: Could we apply Mega's suggestion here of having a separate documentation file to describe the `mode` register and link to each mode-extension from there and vice versa?

EnderShadow: I've done my best at implementing it. Give it a look.

extensions/mode-control-register.md
```diff
@@ -0,0 +1,53 @@
...
+
+| Address Size | Address Mode |
+|--------------|--------------|
+| 16 bit       | word         |
+| 32 bit       | doubleword   |
+| 64 bit       | quadword     |
+
+If the system supports [privilege levels](../privileged-mode/), then `cr17` is only writable when in system privilege mode.
+
+# Recommendations
```
MegaIng, line 47: I would guess this section still best belongs into the individual extensions? Since later extensions make it potentially obsolete. Otherwise the wording should be changed slightly.

extensions/README.md
```diff
@@ -15,8 +15,10 @@
 |     15     | [64 Bit Operations + Registers](./quad-word-operations)   | None          | Mostly Stable     |
 |     16     | [32 Bit Address Space](./32-bit-address-space)            | CP1.14        | Under Development |
 |     17     | Virtual Memory + 32 Bit Paging                            | CP1.16, CP2.2 | Planned           |
-|     32     | 64 Bit Address Space                                      | CP1.15        | Planned           |
-|     33     | Virtual Memory + 64 Bit Paging                            | CP1.32, CP2.2 | Planned           |
+|     32     | [64 Bit Address Space](./64-bit-address-space)            | CP1.15        | Under Development |
+|     33     | Virtual Memory + 64 Bit Paging (48 bit VA)                | CP1.32, CP2.2 | Planned           |
+|     34     | Virtual Memory + 64 Bit Paging (57 bit VA)                | CP1.32, CP2.2 | Planned           |
+|     35     | Virtual Memory + 64 Bit Paging (64 bit VA)                | CP1.32, CP2.2 | Planned           |
```
MegaIng, line 22: Are we sure these are the exact plans we have? If not, I wouldn't list them here.

extensions/64-bit-address-space/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+# Overview
+
+This extension adds a 64-bit addressing mode to the ISA. In particular, this extension focuses on the _Real 64-bit address mode_.
+
+## Real 64-bit Address Mode
+
+A new mode known as _Real 64-bit address mode_ is added, indicated by a value of 4 in `cr17`.
+
+Refer to the [Mode Control Register](../mode-control-register.md) documentation for how real 64-bit addresses behave in relation to other modes.
```
MegaIng: Same as above.

extensions/32-bit-address-space/README.md
```diff
@@ -2,57 +2,14 @@
...
-
-# Recommendations
-
-This extension does not _require_ behaviors beyond what is specified. However, for compatibility with future
-extensions which will require more specifics, the following are recommended:
 
-* The program counter (or equivalent) in the processor itself should always store the sign-
-  extended address when operating in Base mode. This way, nothing needs to be done when
-  entering or leaving real 32-bit address mode from Base mode.
+Refer to the [Mode Control Register](../mode-control-register.md) documentation for how real 32-bit addresses behave in relation to other modes.
```
MegaIng: We could link directly to the correct subsection. `../mode-control-register.md#real-n-bit-address-mode` should be the correct link for that.

extensions/mode-control-register.md
EnderShadow: Why would the privilege-ness of the control register be better in each extension?

MegaIng: Sorry, wrong segment selection. I mean the recommendations section.

extensions/README.md
EnderShadow: You're right, I forgot about the 16 bit paging extension. I'm pretty sure the plan is to stay have similar paging modes as x86 and x86-64. The main difference is that we're skipping 32 bit PAE and we're defining 64 bit paging which goes all the way to the 64 bit address space limit. The reason for these different levels is runtime performance. Each increase in size requires an additional page table layer.

16 bit: 1 layer
32 bit: 2 layers
64 bit (48): 4 layers
64 bit (56): 5 layers
64 bit (64): 6 layers

MegaIng (dismissed):

extensions/README.md
JKTKops: from discord: https://discord.com/channels/828292123936948244/946806826756882552/1108708220563894332
[this page](https://developer.arm.com/documentation/den0024/a/The-Memory-Management-Unit/Translation-tables-in-ARMv8-A/AArch64-descriptor-format) and the next one describe page tables in aarch64. They can be 3 or 4 levels. So I think it would make more sense if 16/32/64 bit paging were incompatible (better pedagogically as it keeps the one you want to focus on simpler). Then the 64 bit ones could be 2/3/4 layers. And we probably only need to support one granule size.

At any rate, that's a planning thing and doesn't affect this PR.

extensions/mode-control-register.md
```diff
@@ -0,0 +1,53 @@
...
+## Base Mode (Real 16-bit Address Mode)
+
+The mode described by [the base isa](../../base-isa.md) is known as the
+Base mode. In that mode, pointers are 16 bits and there is no virtual memory, paging, or memory protection.
+Base mode is indicated by a value of 0 in `cr17`.
+
+## Real n-bit Address Mode
+
+In this mode, all addresses are treated as being `n` bits long. Addresses always refer to their sign extension to the highest supported address size.
+Entering this mode from a lower n-bit address mode _must_ preserve the program counter. Returning to a lower n-bit address mode must preserve the program
```
JKTKops: lower => smaller?

extensions/mode-control-register.md
```diff
@@ -0,0 +1,53 @@
...
+|--------------|--------------|
+| 16 bit       | word         |
+| 32 bit       | doubleword   |
+| 64 bit       | quadword     |
+
+If the system supports [privilege levels](../privileged-mode/), then `cr17` is only writable when in system privilege mode.
+
+# Recommendations
+
+This extension does not _require_ behaviors beyond what is specified. However, for compatibility with future
```
JKTKops:
```suggestion
These extensions do not _require_ behaviors beyond what is specified. However, for compatibility with future
```

JKTKops (dismissed):

extensions/README.md
EnderShadow: I'm not sure how they'd be compatible in the first place aside from using similar page table entry structures.

extensions/mode-control-register.md
```diff
@@ -0,0 +1,53 @@
...
+
+The mode described by [the base isa](../../base-isa.md) is known as the
+Base mode. In that mode, pointers are 16 bits and there is no virtual memory, paging, or memory protection.
+Base mode is indicated by a value of 0 in `cr17`.
+
+## Real n-bit Address Mode
+
+In this mode, all addresses are treated as being `n` bits long. Addresses always refer to their sign extension to the highest supported address size.
+Entering this mode from a smaller n-bit address mode _must_ preserve the program counter. Returning to a smaller n-bit address mode must preserve the program
+counter _if possible_ - if the current program counter interpreted as the lower n-bit Address would not refer to the same location, then the behavior of
```
JKTKops: There were two uses of "lower"... sometimes github's UI for indicating things like this is frustrating :(

JKTKops (dismissed):

extensions/mode-control-register.md
EnderShadow: Looks like there are 3 and I missed one. Fixed now.

JKTKops (approved):

PR #56, opened 2023-05-03
Combine exopc and BM1, add mul-div, add #DE exception

JKTKops:


extensions/expanded-opcodes/README.md
```diff
@@ -32,8 +32,15 @@ This extension adds an expanded instruction format to allow for a larger number
...
 | `0 0000 0010` | `RSBB`          | `A ← B - A - C`                    | `ZNCV` | (1) (2)     |
-| `0 0000 0011` |                 |                                    |        | reserved    |
-| `0 0000 01xx` |                 |                                    |        | reserved    |
+| `0 0000 0011` | `ASR`           | `A ← A >>> B`                      | `C`    | (3) (4)     |
+| `0 0000 0100` | `ROL`           | <code>A ← (A << B) &#124; (A >> -B)</code> | `C` | (3)   |
+| `0 0000 0101` | `ROR`           | <code>A ← (A >> B) &#124; (A << -B)</code> | `C` | (3)   |
+| `0 0000 0110` | `RCL`           | <code>C:A ← (C:A << B) &#124; (C:A >> -B)</code> | `C` | (2) (3) (5) |
+| `0 0000 0111` | `RCR`           | <code>A:C ← (A:C >> B) &#124; (A:C << -B)</code> | `C` | (2) (3) (5) |
+| `0 0000 1000` | `SHL`           | `A ← A << B`                       | `C`   | (3)         |
+| `0 0000 1001` | `SHR`           | `A ← A >> B`                       | `C`   | (3)         |
```
EnderShadow: 1. Why are the zero and negative flags no longer being set? They seem like reasonable flags to set.
2. Why do the normal shifts and rotates set the carry flag? It seems like that would only be well defined for shifts by  1 (and not at all for normal rotates).

extensions/interrupts/README.md
```diff
@@ -49,12 +49,13 @@ The following table specifies which flag is associated with which bit in the `FL
 | Illegal Instruction      | Synchronous  | 2                | Occurs when execution of a reserved, or illegal instruction is attempted.                                          |
 | Memory Alignment Error   | Synchronous  | 3                | Occurs when an attempt to read or write to memory at an unaligned address occurs and is not supported.             |
 | General Protection Fault | Synchronous  | 4                | Occurs when the CPU needs the operating system to handle an unexpected event for stability or consistency reasons. |
-| External Interrupt       | Asynchronous | 5                | Occurs when an external interrupt occurs (eg. keyboard, mouse, etc.).                                              |
+| Divide Error             | Synchronous  | 5                | Occurs when a division by zero is attempted.                                                                       |
```
EnderShadow, line 52: Would it be better to specify a sentinel value for divisions by 0 instead of an interrupt? Interrupts seem like a pretty heavy-handed way of dealing with them.

extensions/interrupts/README.md
```diff
@@ -49,12 +49,13 @@ The following table specifies which flag is associated with which bit in the `FL
...
+| External Interrupt       | Asynchronous | 8                | Occurs when an external interrupt occurs (eg. keyboard, mouse, etc.).                                              |
 
 Synchronous interrupts are interrupts which are triggered directly by the currently running code and must be handled immediately before normal execution can resume.  
 Asynchronous interrupts are interrupts which are triggered by another piece of hardware or indirectly from code and do not need to be handled immediately.
 
-Unused mask/pending bits are reserved for future extensions
+Unused mask/pending bits are reserved for future extensions. External interrupt is at bit 8 to leave a bit of space for other internal exceptions.
```
EnderShadow, line 58: I don't think the specific order for external interrupts matters, does it? If it does, is 2 spaces enough?

extensions/multiply-divide/README.md
```diff
@@ -0,0 +1,50 @@
...
+| `C CCCC CCCC` | NAME     | Operation                                  | Flags  | Comment |
+|---------------|----------|--------------------------------------------|--------|---------|
+| `0 0001 0000` | `UDIV`   | `A ← A / B`                                | None   | (1)     |
+| `0 0001 0001` | `SDIV`   | `A ← A / B`                                | None   | (1)     |
+| `0 0001 0010` | `UREM`   | `A ← A % B`                                | None   | (1) (2) |
+| `0 0001 0011` | `SREM`   | `A ← A % B`                                | None   | (1) (2) |
+| `0 0001 0100` | `UMUL`   | `A ← A × B`                                | `CV`   | (3) (5) |
+| `0 0001 0101` | `SMUL`   | `A ← A × B`                                | `CV`   | (3) (6) |                                          
+| `0 0001 0110` | `UMULH`  | `A ← high(A × B)`                          | None   | (4)     |
+| `0 0001 0111` | `SMULH`  | `A ← high(A × B)`                          | None   | (4)     |
```
EnderShadow: Is there a reason for not defining the Z and N bits for these operations? Also, why define BOTH C and V flags if they're going to be the same?

extensions/expanded-opcodes/README.md
JKTKops: 1. Because other architectures don't do it. Specifically, when implementing OoO processors, it is shockingly useful to rename the flags register _at the bit level_. So the C flag is renamed separately from the V flag, which are renamed separately from the (usually combined) N and Z flags. This way, false flag-carried dependencies can be eliminated. Setting more flags = more dependencies, and checking Z and N after shifts (where the result being Z or N is rarely relevant) is rare. It can always be done by TESTing the result against itself when needed.
2. Every other architecture seems to do it. It's certainly well-defined for all shifts, and the comment explains how. Another way to see it is that the carry flag ends up with the same bit that it _would_ have ended up with if the operation were a rotate through carry - it just doesn't affect the operation otherwise.
I think it's useful for implementing floating point in software, but I'll have to actually do that before I can explain how.

extensions/interrupts/README.md
JKTKops: RISC-V is the only ISA I can find that gives a sentinel value. The sentinel value is -1. ARM _supports_ doing this, but the default action seems to be an exception.

At the end of the day, a program that causes this to happen is erroneous and should be killed. Providing a sentinel value could hide bugs at best and cause very wrong behavior at worst. I think RISC-Vs decision is guided by the fact that they don't want the M extension to depend on interrupts.

extensions/interrupts/README.md
JKTKops: It doesn't, but when this is extended to allow bits 8-50 (for example) to be external interrupts, it would be weird if they were not contiguous. And I actually think 1 might be enough - the only fault that I think is not specified yet is the page fault. But 8 is a nice number and an extra one for safety doesn't hurt.

extensions/multiply-divide/README.md
JKTKops: Mostly matching x86. Also see above about flags.

For ARM, depending on the flavor, either no flags or set, or the Z and N flags are set; C is untouched (or corrupted on old versions) and V is untouched. We could set all of them, I'm not opposed to that. But leaving flags untouched is specifically against part of our rationale.

extensions/expanded-opcodes/README.md
EnderShadow: 1. I'll need to do some of my own research on this, but it seems odd to not set the Z and N flags as well.
2. You do define a way to set carry for shifts and rotate with carry, but that doesn't explain what carry should be for ROL and ROR. Nothing is shifted out for those 2 instructions. Unless you're saying that they effectively become duplicates of the lowest bit?

extensions/multiply-divide/README.md
EnderShadow: I 100% agree that mandating unmentioned flags be untouched isn't something that we should do. I think it makes sense for the Z and N flags to be set with these instructions. I don't think we need to set BOTH C and V. Only setting one of them should be good enough.

EnderShadow: This also brings up the possibility of having ALU instructions which leave ALL the flags in undefined states rather than preserving all of them. We should probably have some sort of way to differentiate them if both of those types of instructions are possible.

extensions/expanded-opcodes/README.md
JKTKops: 1. Alright. For the record I'm implementing this in my build and I'll see if it's useful in the program I want to write.
2. Yes, that's what it says.

extensions/multiply-divide/README.md
JKTKops: I'm not opposed to setting ZNC. I'd like to find out why ARM doesn't set C based on the upper half, but I wouldn't know who to ask. Maybe there's some former ARM developer on stackoverflow lol.

I'm not sure what the benefit of such an ALU instruction would be.

extensions/multiply-divide/README.md
```diff
@@ -0,0 +1,50 @@
...
+1) Non-integral results of divisions are truncated towards zero. Therefore 10 / 3 is 3,
+    and -10 / 3 is -3. The result (quotient or remainder) is zero extended for the unsigned
+    operations, and sign extended for the signed operations.
+2) The remainder of a division is always smaller in magnitude than
+    the divisor, and has the sign of the dividend.
+3) The multiplication operands are treated as having the operation size (as from the `SS` bits).
+    The result will have the next size up. If it cannot be stored (for example, the operand size
+    is `word` and the system does not support [doubleword operations](../double-word-operations)),
+    then the truncation to the operation size is stored. The result of `UMUL` is always zero-extended,
+    and the result of `SMUL` is always sign-extended.
```
EnderShadow: Since we briefly mentioned this scenario earlier, what do we do when the A argument is memory? Do we always truncate to `SS`? Do we always store the full result? Going with the latter, the programmer would need to ensure that the memory address is properly aligned. Although it could also be broken into 2 `SS` sized writes

It might be easier to just get rid of the whole "next size up" part and have separate instructions for the low and high halves.

Another solution might be to specify a prefix which defines the destination for the upper half/remainder.

extensions/expanded-opcodes/README.md
MegaIng: 1. Not setting ZN seems odd to me as well, but I can't really think of a good use case. If we run into something, we could still change this. 
2. Could be mentioned explicitly, but not required IMO

extensions/multiply-divide/README.md
EnderShadow: My preferred solutions to handling the twice as large output is to either have fixed result registers (eg. R0 and R1) or use a prefix to specify a second destination register.

extensions/multiply-divide/README.md
```diff
@@ -0,0 +1,50 @@
...
+An `S` indicates that the operands are treated as signed.
+
+### Opcode table
+
+| `C CCCC CCCC` | NAME     | Operation                                  | Flags  | Comment |
+|---------------|----------|--------------------------------------------|--------|---------|
+| `0 0001 0000` | `UDIV`   | `A ← A / B`                                | None   | (1)     |
+| `0 0001 0001` | `SDIV`   | `A ← A / B`                                | None   | (1)     |
+| `0 0001 0010` | `UREM`   | `A ← A % B`                                | None   | (1) (2) |
+| `0 0001 0011` | `SREM`   | `A ← A % B`                                | None   | (1) (2) |
```
MegaIng: We should mention that these will cause interrupts/undefined behavior on division by zero.

extensions/expanded-opcodes/README.md
```diff
@@ -32,8 +32,15 @@ This extension adds an expanded instruction format to allow for a larger number
...
+| `0 0000 101x` |                 |                                    |        | reserved    |
+| `0 0000 11xx` |                 |                                    |        | reserved    |
 | `0 0000 1xxx` |                 |                                    |        | reserved    |
```
EnderShadow: I think this last line is incorrect and should be removed

EnderShadow (dismissed):

extensions/expanded-opcodes/README.md
MegaIng: Actually, even worse, it's wrong. It implies that `0 0000 1000` is reserved, but actually that is `SHL`

EnderShadow: Well yes, it's incorrect because it's wrong because it was missed in the first commit

EnderShadow (dismissed):

extensions/multiply-divide/README.md
EnderShadow: Thinking about common uses of division and remainder, aren't comparing both against 0 pretty common as well as checking if the remainder is negative (for implementing modulus)? Maybe we want to set the negative flag for signed division and remainder and the zero flag for both divisions and remainders?

MegaIng (changes requested):
Zero division errors still need to be explicitly mentioned.

JKTKops:
This is why I wait for approval from both of you on major PRs :)

Iteration continues...

extensions/multiply-divide/README.md
JKTKops: I agree that that could be useful. Good litmus test for flags: what does ARM do? Unfortunately, it's very hard to find _justifications_ for anything ARM does... I swear it feels like I need a manual for how to read their manuals.

extensions/multiply-divide/README.md
MegaIng: I missed that section completly the first two times I read these PRs 🙃

extensions/multiply-divide/README.md
```diff
@@ -0,0 +1,62 @@
...
+4) The result of a multiplication at size `SS` is at most twice as large.
+    These instructions store only the **upper** half.
+    The result (that is, the upper half of the product) is zero extended for `UMULH` and sign
+    extended for `SMULH`.
+5) The `C` and `V` flags are both set to 0 if the upper half of the multiplication is 0,
+    even if that half would be discarded by the instruction. Otherwise they are set to 1.
+    The values of the `Z` and `N` flags are undefined afterwards.
+6) The `C` and `V` flags are set to 1 if the actual signed value of the product is not equal
+    to the signed truncated result. Otherwise they are set to 0.
+    The values of the `Z` and `N` flags are undefined afterwards.
```
EnderShadow, line 55: Do we need to explicitly say that the `Z` and `N` flags are undefined? Isn't that already specified in the base-isa?

extensions/multiply-divide/README.md
```diff
@@ -0,0 +1,62 @@
...
+| `C CCCC CCCC` | NAME     | Operation                                  | Flags  | Comment |
+|---------------|----------|--------------------------------------------|--------|---------|
+| `0 0001 0000` | `UDIV`   | `A ← A / B`                                | None   | (1)     |
+| `0 0001 0001` | `SDIV`   | `A ← A / B`                                | None   | (1)     |
+| `0 0001 0010` | `UREM`   | `A ← A % B`                                | None   | (1) (2) |
+| `0 0001 0011` | `SREM`   | `A ← A % B`                                | None   | (1) (2) |
+| `0 0001 0100` | `UMUL`   | `A ← low(A × B)`                           | `CV`   | (3) (5) |
+| `0 0001 0101` | `SMUL`   | `A ← low(A × B)`                           | `CV`   | (3) (6) |                                          
+| `0 0001 0110` | `UHMUL`  | `A ← high(A × B)`                          | None   | (4)     |
+| `0 0001 0111` | `SHMUL`  | `A ← high(A × B)`                          | None   | (4)     |
```
EnderShadow: Does `None` for flags mean that the flags are not affected or that they're in an undefined state? Maybe we need to do a re-factor to the rest of the code base to make it explicit which instructions make all flags undefined and which ones don't touch any flags?

extensions/multiply-divide/README.md
EnderShadow: I just looked at the ARM manual and it says that `MUL` updates the `N` and `Z` flags. It also will corrupt or preserve the `C` and `V` flags depending on the version of ARM. Interestingly enough, the HMUL instructions don't affect the flags at all on ARM.

For the division instructions, ARM doesn't touch the conditions.

extensions/multiply-divide/README.md
JKTKops: Clarifications don't hurt.

extensions/multiply-divide/README.md
JKTKops: There are no instructions that make all flags undefined. That wouldn't be useful.

MegaIng (dismissed):

extensions/multiply-divide/README.md
EnderShadow: We should probably be consistent with the rest of the spec then. (even if we do a different PR to change it all to None instead of being blank)

extensions/expanded-opcodes/README.md
EnderShadow:
```suggestion
| `0 0000 0011` | `ASR`           | `A ← A >>> B`                      | `CZN`    | (3) (4)     |
| `0 0000 0100` | `ROL`           | <code>A ← (A << B) &#124; (A >> -B)</code> | `CZN` | (3)   |
| `0 0000 0101` | `ROR`           | <code>A ← (A >> B) &#124; (A << -B)</code> | `CZN` | (3)   |
| `0 0000 0110` | `RCL`           | <code>C:A ← (C:A << B) &#124; (C:A >> -B)</code> | `CZN` | (2) (3) (5) |
| `0 0000 0111` | `RCR`           | <code>A:C ← (A:C >> B) &#124; (A:C << -B)</code> | `CZN` | (2) (3) (5) |
| `0 0000 1000` | `SHL`           | `A ← A << B`                       | `CZN`   | (3)         |
| `0 0000 1001` | `SHR`           | `A ← A >> B`                       | `CZN`   | (3)         |
```

My justification for this suggestion is that, while Z and N may not be used, every other ALU instruction sets these flags so it would add functionality while also potentially simplifying hardware.

extensions/multiply-divide/README.md
EnderShadow:
```suggestion
| `0 0001 0000` | `UDIV`   | `A ← A / B`                                | `ZN`   | (1)     |
| `0 0001 0001` | `SDIV`   | `A ← A / B`                                | `ZN`   | (1)     |
| `0 0001 0010` | `UREM`   | `A ← A % B`                                | `ZN`   | (1) (2) |
| `0 0001 0011` | `SREM`   | `A ← A % B`                                | `ZN`   | (1) (2) |
| `0 0001 0100` | `UMUL`   | `A ← low(A × B)`                           | `CZN`  | (3) (5) |
| `0 0001 0101` | `SMUL`   | `A ← low(A × B)`                           | `CZN`  | (3) (6) |                                          
| `0 0001 0110` | `UHMUL`  | `A ← high(A × B)`                          | `ZN`   | (4)     |
| `0 0001 0111` | `SHMUL`  | `A ← high(A × B)`                          | `ZN`   | (4)     |                                          
```
The main justification for this is that it can simplify the ALU as mentioned above. There are a few reasons these might be useful anyway though.

- SREM setting `N` will allow for a more compact modulus instruction.
- Setting both `C` and `V` to the same value for UMUL and SMUL seems redundant, but I don't feel too strongly about it.

extensions/multiply-divide/README.md
EnderShadow: Suggested changes are below as this is considered out-of-date by github

extensions/multiply-divide/README.md
JKTKops: (this change was discussed on Discord beforehand).

EnderShadow (approved):

MegaIng (approved):

PR #61, opened 2023-07-19
Added abbreviations to extensions and features.

MegaIng:
I think it would be helpful to have these for the GNU toolchain, and potentially also as commandline options. A few of these are already in common usage on discord, others I invented. I am not 100% sold on all of these, so suggestions are welcome.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -1,34 +1,34 @@
 # Extensions
 
-| CPUID1 bit | Extension                                                 | Dependencies  | State             |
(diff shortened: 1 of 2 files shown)


EnderShadow (dismissed):

EnderShadow:
They seem fine to me. The only potential changes might be with the VM ones changeing to PG for paging

EnderShadow (approved):

JKTKops (approved):
LGTM. BYTE instead of HW is maybe inconsistent, but not bad.

PR #65, opened 2023-08-08
Clarify SAF

JKTKops:
As discussed in Discord, + some other minor clarifications and cross-extension links.


MegaIng (approved):

extensions/stack-and-functions/README.md
```diff
@@ -47,7 +47,7 @@ The following opcodes are now defined.
 | Symbol | Meaning                                    |
```
EnderShadow, line 47: Shouldn't we also change the table above to use `AAA` and `BBB`?

EnderShadow (changes requested):

EnderShadow (approved):

PR #67, opened 2023-08-21
Interrupt Rework 2: Electric Boogaloo

EnderShadow:
Based on our conversation in discord, this PR does the following:
1) merges the external and timer interrupt
2) removes the `INT_PENDING` and `INT_MASK` CRs

This makes it so that the CPU handles all synchronous exceptions on it's own and relies on a PIC for all external interrupts. This implies that the PIC is controlled through MMIO and starts off in a disabled state.

### DIFF
--- a/extensions/32-bit-address-space/README.md
+++ b/extensions/32-bit-address-space/README.md
@@ -10,6 +10,6 @@ This extension adds a 32-bit addressing mode to the ISA. In particular, this ext
 
 ## Real 32-bit Address Mode
 
-A new mode known as _Real 32-bit address mode_ is added, indicated by a value of 2 in `cr17`.
+A new mode known as _Real 32-bit address mode_ is added, indicated by a value of 2 in `cr15`.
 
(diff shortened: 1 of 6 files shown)


JKTKops:
Sparked by discussion beginning here:
https://discord.com/channels/828292123936948244/946806826756882552/1142980237932626011

TL;DR the existing interrupt mask/pending system was a bit half baked and did some critical jobs of an interrupt controller but not others, leading to a majority of situations where it can't be used at all and PIC would've been necessary anyway.

We want to better understand what other ISAs do (both PIC-based and ISA-based) before settling on particular changes.

EnderShadow:
Another thing we might want to change is how external interrupts are triggered. Currently they're edge triggered, but since we're relying on a PIC, it might be better to change it to level triggered?
https://github.com/ETC-A/etca-spec/blob/2cb33131c3d921dfaea4ba7abac6b56ad3fcf4a6/extensions/interrupts/README.md?plain=1#L79

PR #68, opened 2023-08-23
Clarify mo1 and pointers

JKTKops:
I'm making these edits through the github UI so not quite ready yet

### DIFF
--- a/extensions/memory-operands-1/README.md
+++ b/extensions/memory-operands-1/README.md
@@ -69,34 +69,42 @@ read or write in which case it is _unspecified_ behavior.
 
 The `D` bit is the less significant `MM` bit, and refers to the "direction" of the operation.
 
-Let `A` refer to the value of the register specified by `AAA`. Let `M` refer to the value in the memory
-location denoted by the address as described above. Then all operations using this format have the following operation:
+Let `A` refer to the value of the register specified by `AAA`. Let `M` refer to the address encoded
+as described above, and let `[M]` be the value at that location. 
+Then all operations using this format have the following operation:
 ```
 if D == 0:
   OP_L ← A
-  OP_R ← M
+  if operation == LEA:
+    OP_R ← M
+  else:
+    OP_R ← [M]
 else:
   if dst_is_input(operation):
-    OP_L ← M
+    OP_L ← [M]
   OP_R ← A
 RES ← operation(OP_L, OP_R)
 if stores_result(operation):
   if D == 0:
     A ← RES
   else:
-    M ← RES
+    [M] ← RES
 ```
 In plain language; when `D` is 0, the `dst` operand is a register and the `src` operand is a memory location.
-When `D` is 1, these are swapped. In all cases, the value in the memory location must be loaded from memory.
-When the `dst` operation is a memory location, the result of the operation must also be stored there.
+When `D` is 1, these are swapped. Usually, the value in the memory location must be loaded from memory.
+When the `dst` operation is a memory location, the result of the operation must also usually be stored there. However:
 
-If an operation uses an `ABM` byte to encode two operands, but only treats one of them as an input, then this
-behavior applies likewise to memory operations; memory must not be (visibly) loaded if the memory operand is not an input
-(e.g., as is the case for `mov m, r/i`). If the operation does not store its result, then memory must not be modified.
+- If an operation uses an `ABM` byte to encode two operands, but does not treat the memory one as input,
+    memory must not be (visibly) loaded. Note that this means MMIO address cannot be read from, even
+    if the result of the read is immediately discarded.
+- If the operation is `LEA`, memory must likewise not be (visibly) loaded. `LEA` is interested in the address itself.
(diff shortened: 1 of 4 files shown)


EnderShadow (dismissed):

extensions/memory-operands-1/README.md
```diff
@@ -69,34 +69,42 @@ read or write in which case it is _unspecified_ behavior.
...
+    if the result of the read is immediately discarded.
+- If the operation is `LEA`, memory must likewise not be (visibly) loaded. `LEA` is interested in the address itself.
+- If the operation does not store its result (e.g. `cmp m,r`), then memory must not be (visibly) written to.
+    This also includes MMIO addresses.
 
 The instruction's Operand Size (usually from the instruction `SS` bits, but NOT from the `SIB.S` bits) informs the size of the read or writes. These reads
 and writes must behave, with regards to how memory is read and modified, the same as the `load` and `store` instructions
-to the same address, with the same operand size.
+to the same address, with the same operand size. When [multiple memory access instructions](../../features/multiple-memory-access-instructions/README.md)
+is available, the size of a nested read or write therefore matches the current Address Size attribute.
```
MegaIng, line 105: I think we should explain the various cases here a bit more (it took me quite a while to realize what you mean with "nested *write*"). Maybe it doesn't belong right here, and instead should be explained over on the feature page in detail?

(It also doesn't really follow at all from the above pseudocode, but adding it there would make the code way to complicated)

extensions/memory-operands-1/README.md
```diff
@@ -69,34 +69,42 @@ read or write in which case it is _unspecified_ behavior.
...
 
-If an operation uses an `ABM` byte to encode two operands, but only treats one of them as an input, then this
-behavior applies likewise to memory operations; memory must not be (visibly) loaded if the memory operand is not an input
-(e.g., as is the case for `mov m, r/i`). If the operation does not store its result, then memory must not be modified.
+- If an operation uses an `ABM` byte to encode two operands, but does not treat the memory one as input,
+    memory must not be (visibly) loaded. Note that this means MMIO address cannot be read from, even
+    if the result of the read is immediately discarded.
+- If the operation is `LEA`, memory must likewise not be (visibly) loaded. `LEA` is interested in the address itself.
+- If the operation does not store its result (e.g. `cmp m,r`), then memory must not be (visibly) written to.
+    This also includes MMIO addresses.
```
MegaIng: (very redundant IMO: You can't do any valid store, visible or invisible, for operations that don't have a result)

extensions/memory-operands-1/README.md
```diff
@@ -113,16 +121,20 @@ and may change in the future when we are better able to evaluate the waste.
...
+| LEA r, m | `00SS1110` | `ABM` | Loads the address specified by the second operand into the register specified by the first. Note that the address _itself_ is stored in the register, not the contents of memory at that address. The result stored in the register must respect the operand size attribute (and is sign extended). The computed address must respect the address mode.
 
 `LEA` means "load effective address." If the first operand is not a register, or the second is not a memory location, the instruction must be treated as an
 _illegal_ instruction. Note that as per the
 [Full Immediates extension](../full-immediates/README.md) in the specific case of `MM=01, ABM.regB=01x`
 (the "full immediate" operand modes) this opcode corresponds to `READCR`, not `LEA`.
 
-Note that LEA shares an opcode with `READCR` (aka `mfcr`). `LEA` has no immediate mode, and `READCR` has
+Note also that LEA shares an opcode with `READCR` (aka `mfcr`). `LEA` has no immediate mode, and `READCR` has
```
MegaIng: These two sentences overlap quite a bit, but the second one acts as if it introduces the overlap with `READCR`.

MegaIng (changes requested):

extensions/memory-operands-1/README.md
EnderShadow: I agree that the note on cmp is unneeded.

extensions/memory-operands-1/README.md
EnderShadow: I agree with Mega

JKTKops:
Let's close this one, either by merging something or shutting it down.

I am replying to some old comments to facilitate necessary discussion.

extensions/memory-operands-1/README.md
JKTKops: OK, let's delete the whole paragraph.

extensions/memory-operands-1/README.md
JKTKops: I think it should be explained under MMAI in that case.

However MMAI is not a feature that I have any interest in maintaining (I do not use it in my code or my builds as it seems needlessly complex, and was only really added to the spec because "hey, we can encode it").

Would either one of you like to take up explaining the various cases here?

JKTKops:
With respect to MMAI, it's currently definitely underspecified (I just looked at implementing it in Mega's sim and then abandoned that). That no one has noticed this in 2+ years suggests to me that no one is using it, so if we just want to make those encodings illegal again, it won't break anything.

PR #69, opened 2023-08-24
CR Enhancement

EnderShadow:


MegaIng (dismissed):

MegaIng (dismissed):

JKTKops (changes requested):
I think the term is "vendor" specific extensions. If we just wanted _implementation_ specific extensions, we'd only need 1 bit (indicating that "some implementation-specific things are supported").

Also, this might be the right time to specify that extension we were considering that would add a CR which reads a processor name as a string based on the value of the destination register before the read.

EnderShadow:
> I think the term is "vendor" specific extensions. If we just wanted _implementation_ specific extensions, we'd only need 1 bit (indicating that "some implementation-specific things are supported").
> 
> Also, this might be the right time to specify that extension we were considering that would add a CR which reads a processor name as a string based on the value of the destination register before the read.

Fixed it so it says vendor.

I'm not entirely on board with the behavior changing based on the destination register, but I'm not strictly against it either.

MegaIng (approved):
This PR is done if we don't want to add the behavior changes for the CR registers here. I am not really opposed to that, but IMO it would be cleaner to just reserve more CRs for this kind of stuff. Not like we are running out.

EnderShadow:
> This PR is done if we don't want to add the behavior changes for the CR registers here. I am not really opposed to that, but IMO it would be cleaner to just reserve more CRs for this kind of stuff. Not like we are running out.

I agree, I think it would be cleaner to reserve more CRs I'm not sure the best place to put them though.

JKTKops (approved):
Seems we decided not to make the other change to CRs and this has been pending for a long time, let's get this one closed.

PR #71, opened 2024-02-18
Draft BM1 extension and move rotate with carry into it

EnderShadow:
The only instruction missing from x86's BMI1 is `BEXTR`, but that requires 3 arguments. Instead of `BEXTR`, I've included `BZHI` which from BMI2

### DIFF
--- a/extensions/bit-manipulation-1/README.md
+++ b/extensions/bit-manipulation-1/README.md
@@ -26,8 +26,8 @@ These instructions are in the expanded calculation opcode section of instruction
 | `0 0000 1111` | `ANDN`    | <code>A ← ~A &#38 B</code>                             | `ZN`  |             |
 | `0 0001 1000` | `LSB`     | <code>A ← B &#38 -B</code>                             | `ZN`  |             |
 | `0 0001 1001` | `LSMSK`   | <code>A ← B ^ (B - 1)</code>                           | `ZN`  |             |
-| `0 0001 1010` | `RLSB`    | <code>A ← B &#38 (B - 1)</code>                        | `ZN`  |             |
-| `0 0001 1011` | `ZHIB`    | <code>A ← A &#38 ((1 << B) - 1)</code>                 | `ZN`  |             |
+| `0 0001 1010` | `RLSB`    | <code>A ← B ;&#38 (B - 1)</code>                       | `ZN`  |             |
+| `0 0001 1011` | `ZHIB`    | <code>A ← A ;&#38 ((1 << B) - 1)</code>                | `ZN`  |             |
 
 1) C in the operation column refers to the carry flag.
 2) Only the `M` least significant bits of `B` (or `-B` where applicable) are used for a shift amount.
--- a/extensions/bit-manipulation-1/README.md
+++ b/extensions/bit-manipulation-1/README.md
@@ -59,7 +59,7 @@ int64_t grev(int64_t a, int b, int ss)
 ```
 
 - Byte swap can be formulated in terms of GREV where B=24 (for `SS = 10`).
-- Bit Reverse can be formulated in terms of GREV where B=31 (for `SS = 11`).
+- Bit Reverse can be formulated in terms of GREV where B=31 (for `SS = 10`).
 
 6) Counts the number of `0` bits before the first `1` bit
 7) Counts the number of `0` bits after the last `1` bit
--- a/extensions/bit-manipulation-1/README.md
+++ b/extensions/bit-manipulation-1/README.md
@@ -27,7 +27,7 @@ These instructions are in the expanded calculation opcode section of instruction
 | `0 0001 1000` | `LSB`     | <code>A ← B &#38; -B</code>                            | `ZN`  | (9)         |
 | `0 0001 1001` | `LSMSK`   | <code>A ← B ^ (B - 1)</code>                           | `ZNC` | (6) (10)    |
 | `0 0001 1010` | `RLSB`    | <code>A ← B &#38; (B - 1)</code>                       | `ZNC` | (6) (11)    |
-| `0 0001 1011` | `ZHIB`    | <code>A ← A &#38; ((1 << B) - 1)</code>                | `ZN`  | (12)        |
+| `0 0001 1011` | `ZHIB`    | <code>A ← A &#38; ((1 << B[7:0]) - 1)</code>           | `ZNC` | (12)        |
 
 1) C in the operation column refers to the carry flag.
 2) The `-1` in the operation equivalently indicates a shift amount of one less than the operation size.
@@ -70,14 +70,16 @@ int64_t grev(int64_t a, int b, int ss)
         grev   A,-1
         clz    A,A
 ```
-9) Notes 9, 10, 11, and 12 give examples of the remaining operations; however the behavior
+9) Notes 9, 10, and 11 give examples of the remaining operations; however the behavior
     of these operations is fully specified in the table and comment (6).
    `lsb` isolates the least significant `1` bit of `B`. For example, `lsbx r0, 0xFFA0`
     will result in `0x0020` in `rx0`.
-11) Get a mask of all bits up to and including the least significant `1` bit of `B`.
+10) Get a mask of all bits up to and including the least significant `1` bit of `B`.
     If there is no such bit, the mask is `-1`. For example, `lsmskx r0, 0xFFA0`
     will result in `0x003F` in `rx0`.
-12) Reset (set to `0`) the least significant `1` bit of `B`. For example,
+11) Reset (set to `0`) the least significant `1` bit of `B`. For example,
     `rlsbx r0, 0xFFA0` will result in `0xFF80` in `rx0`.
-13) Zero out the bits of `A` at higher bit indices at least `B`. For example,
-    if `0xABCD` is in `rx0`, then `zhibx r0, 7` results in `0x004D` in `rx0`.
+12) Zero out the bits of `A` at bit indices which are at least `B`. Only the bottom byte
+    of `B` is considered. If `B` exceeds the operation width minus 1, then `A` is unchanged
+    and the `C` flag is set.
+    For example, if `0xABCD` is in `rx0`, then `zhibx r0, 7` results in `0x004D` in `rx0`.
--- a/extensions/bit-manipulation-1/README.md
+++ b/extensions/bit-manipulation-1/README.md
@@ -16,8 +16,8 @@ These instructions are in the expanded calculation opcode section of instruction
 
 | `C CCCC CCCC` | NAME      | Operation                                              | Flags | Comment     |
 |---------------|-----------|--------------------------------------------------------|-------|-------------|
-| `0 0000 1000` | `RCL`     | <code>C:A ← (C:B << 1) &#124; (C:B >> -1)</code>       | `ZNC` | (1) (2) (3) |
-| `0 0000 1001` | `RCR`     | <code>A:C ← (B:C >> 1) &#124; (B:C << -1)</code>       | `ZNC` | (1) (2) (3) |
+| `0 0000 1000` | `RCL`     | <code>C:A ← (C:B << 1) &#124; C</code>                 | `ZNC` | (1) (2) (3) |
+| `0 0000 1001` | `RCR`     | <code>A:C ← (B:C >> 1) &#124; (C << -1)</code>         | `ZNC` | (1) (2) (3) |
 | `0 0000 1010` | `POPCNT`  | <code>A ← POPCNT(B)</code>                             | `Z`   | (4)         |
 | `0 0000 1011` | `GREV`    | <code>A ← GREV(A, B)</code>                            | `ZN`  | (5)         |
 | `0 0000 1100` | `CTZ`     | <code>A ← CTZ(B)</code>                                | `ZC`  | (6) (7)     |
--- a/extensions/bit-manipulation-1/README.md
+++ b/extensions/bit-manipulation-1/README.md
@@ -24,6 +24,10 @@ These instructions are in the expanded calculation opcode section of instruction
 | `0 0000 1101` | `CLZ`     | Counts the number of `0` bits after the last `1` bit   | `ZN`  |             |
 | `0 0000 1110` | `NOT`     | <code>A ← ~B</code>                                    | `ZN`  |             |
 | `0 0000 1111` | `ANDN`    | <code>A ← ~A &#38 B</code>                             | `ZN`  |             |
+| `0 0001 1000` | `LSB`     | <code>A ← B &#38 -B</code>                             | `ZN`  |             |
+| `0 0001 1001` | `LSMSK`   | <code>A ← B ^ (B - 1)</code>                           | `ZN`  |             |
+| `0 0001 1010` | `RLSB`    | <code>A ← B &#38 (B - 1)</code>                        | `ZN`  |             |
+| `0 0001 1011` | `ZHIB`    | <code>A ← A &#38 ((1 << B) - 1)</code>                 | `ZN`  |             |
 
 1) C in the operation column refers to the carry flag.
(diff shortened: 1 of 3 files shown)


MegaIng:
Also discussed here: https://discord.com/channels/828292123936948244/946806826756882552/1208923377763360769

extensions/bit-manipulation-1/README.md
```diff
@@ -0,0 +1,48 @@
...
+| `0 0000 1010` | `POPCNT`  | Counts the number of set bits in the `B` input         | `ZN`  | (4)         |
+| `0 0000 1011` | `BITSWAP` | Swaps all the bits based on the `SS` bits              | `ZN`  | (5)         |
+| `0 0000 1100` | `CTZ`     | Counts the number of `0` bits before the first `1` bit | `ZN`  |             |
+| `0 0000 1101` | `CLZ`     | Counts the number of `0` bits after the last `1` bit   | `ZN`  |             |
+| `0 0000 1110` | `NOT`     | <code>A ← ~B</code>                                    | `ZN`  |             |
+| `0 0000 1111` | `ANDN`    | <code>A ← ~A &#38 B</code>                             | `ZN`  |             |
+| `0 0001 1000` | `LSB`     | <code>A ← B &#38 -B</code>                             | `ZN`  |             |
+| `0 0001 1001` | `LSMSK`   | <code>A ← B ^ (B - 1)</code>                           | `ZN`  |             |
+| `0 0001 1010` | `RLSB`    | <code>A ← B &#38 (B - 1)</code>                        | `ZN`  |             |
+| `0 0001 1011` | `ZHIB`    | <code>A ← A &#38 ((1 << B) - 1)</code>                 | `ZN`  |             |
```
MegaIng: These http escapes aren't correct, the `;` is missing, see the rich preview on github.
```suggestion
| `0 0000 1111` | `ANDN`    | <code>A ← ~A &#38; B</code>                             | `ZN`  |             |
| `0 0001 1000` | `LSB`     | <code>A ← B &#38; -B</code>                             | `ZN`  |             |
| `0 0001 1001` | `LSMSK`   | <code>A ← B ^ (B - 1)</code>                           | `ZN`  |             |
| `0 0001 1010` | `RLSB`    | <code>A ← B &#38; (B - 1)</code>                        | `ZN`  |             |
| `0 0001 1011` | `ZHIB`    | <code>A ← A &#38; ((1 << B) - 1)</code>                 | `ZN`  |             |
```

extensions/bit-manipulation-1/README.md
```diff
@@ -0,0 +1,65 @@
...
+5) Performs the generalized swap operation on `A` based on the value in `B`. It acts as follows
+```
+int64_t grev(int64_t a, int b, int ss)
+{
+    if (b &  1)           a = ((a & 0x55555555) <<  1) | ((a & 0xAAAAAAAA) >>  1);
+    if (b &  2)           a = ((a & 0x33333333) <<  2) | ((a & 0xCCCCCCCC) >>  2);
+    if (b &  4)           a = ((a & 0x0F0F0F0F) <<  4) | ((a & 0xF0F0F0F0) >>  4);
+    if (b &  8 && ss > 0) a = ((a & 0x00FF00FF) <<  8) | ((a & 0xFF00FF00) >>  8);
+    if (b & 16 && ss > 1) a = ((a & 0x0000FFFF) << 16) | ((a & 0xFFFF0000) >> 16);
+    if (b & 32 && ss > 2) a = ((a & 0x0000FFFF) << 32) | ((a & 0xFFFF0000) >> 32);
```
MegaIng:
```suggestion
    if (b &  1)           a = ((a & 0x5555555555555555) <<  1) | ((a & 0xAAAAAAAAAAAAAAAA) >>  1);
    if (b &  2)           a = ((a & 0x3333333333333333) <<  2) | ((a & 0xCCCCCCCCCCCCCCCC) >>  2);
    if (b &  4)           a = ((a & 0x0F0F0F0F0F0F0F0F) <<  4) | ((a & 0xF0F0F0F0F0F0F0F0) >>  4);
    if (b &  8 && ss > 0) a = ((a & 0x00FF00FF00FF00FF) <<  8) | ((a & 0xFF00FF00FF00FF00) >>  8);
    if (b & 16 && ss > 1) a = ((a & 0x0000FFFF0000FFFF) << 16) | ((a & 0xFFFF0000FFFF0000) >> 16);
    if (b & 32 && ss > 2) a = ((a & 0x00000000FFFFFFFF) << 32) | ((a & 0xFFFFFFFF00000000) >> 32);
```

extensions/bit-manipulation-1/README.md
```diff
@@ -0,0 +1,65 @@
...
+    if `rh0` is `0x80`, then after `rolh rh0,1`, the carry flag will be set and `rh0` will be 1.
+    If the shift amount is 0, the carry flag is undefined.
+    In all cases, the Z and N flags are set according to the result written.
+    For example, if `rh0` is `0x80`, then after `rclh rh0,1`, the carry flag will be set,
+    `rh0` will be 0, the negative flag will be cleared, and the zero flag will be set.
+3) The notation C:A represents combining the carry flag with the value in register `A`. If the operation
+    size is `half`, the value is `C AAAA AAAA`, a 9-bit value. The `RCL` instruction rotates this 9-bit
+    value left. The `RCR` instruction rotates the value `AAAA AAAA C` to the right. The operation
+    is analogous for other operation sizes.
+4) Counts the number of set bits in `B` as if it was zero extended based on the `SS` bits.
```
JKTKops:
```suggestion
4) Counts the number of set bits in `B` as if it were zero extended based on the `SS` bits.
```
grammar

extensions/bit-manipulation-1/README.md
```diff
@@ -0,0 +1,65 @@
...
+| `0 0000 1001` | `RCR`     | <code>A:C ← (A:C >> B) &#124; (A:C << -B)</code>       | `ZNC` | (1) (2) (3) |
+| `0 0000 1010` | `POPCNT`  | <code>A ← POPCNT(B)</code>                             | `ZN`  | (4)         |
+| `0 0000 1011` | `GREV`    | <code>A ← GREV(A, B)</code>                            | `ZN`  | (5)         |
+| `0 0000 1100` | `CTZ`     | <code>A ← CTZ(B)</code>                                | `ZN`  | (6)         |
+| `0 0000 1101` | `CLZ`     | <code>A ← CLZ(B)</code>                                | `ZN`  | (7)         |
+| `0 0000 1110` | `NOT`     | <code>A ← ~B</code>                                    | `ZN`  |             |
+| `0 0000 1111` | `ANDN`    | <code>A ← ~A &#38; B</code>                            | `ZN`  |             |
+| `0 0001 1000` | `LSB`     | <code>A ← B &#38; -B</code>                            | `ZN`  |             |
+| `0 0001 1001` | `LSMSK`   | <code>A ← B ^ (B - 1)</code>                           | `ZN`  |             |
+| `0 0001 1010` | `RLSB`    | <code>A ← B &#38; (B - 1)</code>                       | `ZN`  |             |
```
JKTKops: Let's include a description of these last 4 operations. I don't think we should expect the average person reading this document to already be familiar with what these expressions do.

A page on the wiki motivating these instructions might be a good idea as well? Honestly it's a bit difficult to motivate `rcl`/`rcr` for shift amounts other than 1. Do we know of any use-cases for that? It's a little bit awkward, but I might not be opposed to only specifying support for rcl/rcr by 1 with a feature or possibly later extension specifying support for other amounts.

EDIT: I did some research on this. Amounts other than 1 incur slowdowns on x86 and according to StackOverflow are "rarely used, if ever."

Further edit: If we go with that, we should perhaps also mention for the benefit of people who haven't noticed yet, that `rcl x,1` is the same operation as `adc x, x` but doesn't set `V`.

JKTKops (dismissed):

extensions/bit-manipulation-1/README.md
EnderShadow: I have no problems with making rcr and rcl limited to shifts by 1. We could also just remove rcl in that case because of the adc overlap

MegaIng: If we turn `rcr`/`rcl` into unary operations, it is once again similar to the XOR/NOT operation, so I wouldn't be opposed to keeping it. For symmetry with everything else in the assembly we could have the pseudo instructions `rcr, %rx0, 1` which are only valid with immediate=1.

EnderShadow: We could also make it so that `rcr, a, b` performs the operation on `b` and stores it in `a`

JKTKops: There are other options, like the original idea of leaving values other than 1 to a feature or something.

I can't even figure out if ARM has rotate through carry instructions. Most old microcontrollers don't seem to support values other than 1. x86 added support for that via microcoding at some point but I can't find any use cases of it. I think I actually link Ender's idea here. The only downside is if using these instructions to implement a bigint or softfloat library, I think you might actually want the operation `rcr [a], 1` to be one instruction, which isn't possible for a version that isn't in-place. That's a small cost though and I'm not even sure it applies.

EnderShadow: In that case, I'm fine with fixing the second argument at 1 unless a feature is implemented if we don't want to do my above suggestion.

EnderShadow: It looks like ARM has `rcr dst, src` which does rotate right with carry by 1. It's called `rrx` (rotate right with extend). They don't have `rcl`, but that's because they just use `adc dst, src, src` to emulate it.
https://developer.arm.com/documentation/dui0489/i/arm-and-thumb-instructions/rrx

JKTKops: OK. I'm down to replace the instructions specified here with src->dst rotates by 1. There's a clear distinction between such an `rcl` and an `adc` so it can stay. The immediate modes seem completely useless but it would be unnecessary complication to reserve them, I think. And, technically, `rcl r, 15` would be a 1-instruction (3-byte) way to put 30 in a register without FI. Not that that sounds useful.

extensions/bit-manipulation-1/README.md
```diff
@@ -0,0 +1,65 @@
...
+|---------------|-----------|--------------------------------------------------------|-------|-------------|
+| `0 0000 1000` | `RCL`     | <code>C:A ← (C:A << B) &#124; (C:A >> -B)</code>       | `ZNC` | (1) (2) (3) |
+| `0 0000 1001` | `RCR`     | <code>A:C ← (A:C >> B) &#124; (A:C << -B)</code>       | `ZNC` | (1) (2) (3) |
+| `0 0000 1010` | `POPCNT`  | <code>A ← POPCNT(B)</code>                             | `ZN`  | (4)         |
+| `0 0000 1011` | `GREV`    | <code>A ← GREV(A, B)</code>                            | `ZN`  | (5)         |
+| `0 0000 1100` | `CTZ`     | <code>A ← CTZ(B)</code>                                | `ZN`  | (6)         |
+| `0 0000 1101` | `CLZ`     | <code>A ← CLZ(B)</code>                                | `ZN`  | (7)         |
+| `0 0000 1110` | `NOT`     | <code>A ← ~B</code>                                    | `ZN`  |             |
+| `0 0000 1111` | `ANDN`    | <code>A ← ~A &#38; B</code>                            | `ZN`  |             |
+| `0 0001 1000` | `LSB`     | <code>A ← B &#38; -B</code>                            | `ZN`  |             |
```
JKTKops: Why the jump from `0F` to `18`?

MegaIng: Because the MD extension takes up that space. Maybe we could have a more consistent pattern here, but IMO it's fine.

JKTKops: Oh, of course.

extensions/bit-manipulation-1/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+    return sign_extend(a, ss);
+}
+```
+
+- Byte swap can be formulated in terms of GREV where B=24 (for `SS = 10`).
+- Bit Reverse can be formulated in terms of GREV where B=31 (for `SS = 10`).
+
+6) Counts the number of `0` bits before the first `1` bit
+7) Counts the number of `0` bits after the last `1` bit
```
JKTKops: In addition to the requested descriptions of the last 4 operations, these also need to say what happens if there are no set bits in the input.

extensions/bit-manipulation-1/README.md
```diff
@@ -16,34 +16,31 @@ These instructions are in the expanded calculation opcode section of instruction
 
 | `C CCCC CCCC` | NAME      | Operation                                              | Flags | Comment     |
 |---------------|-----------|--------------------------------------------------------|-------|-------------|
-| `0 0000 1000` | `RCL`     | <code>C:A ← (C:A << B) &#124; (C:A >> -B)</code>       | `ZNC` | (1) (2) (3) |
-| `0 0000 1001` | `RCR`     | <code>A:C ← (A:C >> B) &#124; (A:C << -B)</code>       | `ZNC` | (1) (2) (3) |
+| `0 0000 1000` | `RCL`     | <code>C:A ← (C:B << 1) &#124; (C:B >> -1)</code>       | `ZNC` | (1) (2) (3) |
+| `0 0000 1001` | `RCR`     | <code>A:C ← (B:C >> 1) &#124; (B:C << -1)</code>       | `ZNC` | (1) (2) (3) |
```
MegaIng: I think now that it's fixed shift, it would be better to write `B ← (B << 1) | C; C = msb(B)` or something like that.

MegaIng (changes requested):

extensions/bit-manipulation-1/README.md
```diff
@@ -0,0 +1,83 @@
...
+    of these operations is fully specified in the table and comment (6).
+   `lsb` isolates the least significant `1` bit of `B`. For example, `lsbx r0, 0xFFA0`
+    will result in `0x0020` in `rx0`.
+11) Get a mask of all bits up to and including the least significant `1` bit of `B`.
+    If there is no such bit, the mask is `-1`. For example, `lsmskx r0, 0xFFA0`
+    will result in `0x003F` in `rx0`.
+12) Reset (set to `0`) the least significant `1` bit of `B`. For example,
+    `rlsbx r0, 0xFFA0` will result in `0xFF80` in `rx0`.
+13) Zero out the bits of `A` at higher bit indices at least `B`. For example,
+    if `0xABCD` is in `rx0`, then `zhibx r0, 7` results in `0x004D` in `rx0`.
```
MegaIng: We should probably explicitly mentioned what the behavior should be if `B >= reg_width`

extensions/bit-manipulation-1/README.md
JKTKops: The `C:B` notation covers setting the `C` flag (and is explained again in a comment) so I think that that's OK. However using `| C` for the bottom bit is clearly better. Making an update shortly.

MegaIng: Yes, but IMO it would better to not require the comment explaining a tbh. slightly arcane syntax for a situations where it isn't required. Saying "`msb(B)` refers to the most significant bit of `B` for the correct width" is IMO clearer and easier to understand.

extensions/bit-manipulation-1/README.md
JKTKops: Ah... good point. My first draft microcode would actually have missed this and would've taken `B mod reg_width`.

JKTKops: Hmm, actually, this is a missing point in the spec, because the behavior of `<<` in this case is ambiguous.

x86 (1) takes only the bottom 8 bits of `B`, (2) treats indices above `operation_width-1` as being equal to `operation_width-1` (i.e. no change to input), AND (3) set `C`.

That seems reasonable for our case as well, but there are other options.

MegaIng (approved):

PR #72, opened 2024-03-16
Add table layout which matches the binary layout for expanded opcodes

EnderShadow:


MegaIng (dismissed):

JKTKops:
What's the point of the "alternate table layout"s? I think they're confusing - I didn't even notice what was different for a while because I was looking for the table itself to be rearranged, not the bit patterns. And then it looks at first like the singular bit has been moved, rather than the bits being regrouped.

EnderShadow:
> What's the point of the "alternate table layout"s? I think they're confusing - I didn't even notice what was different for a while because I was looking for the table itself to be rearranged, not the bit patterns. And then it looks at first like the singular bit has been moved, rather than the bits being regrouped.

The point of the layout was that the bit format matches the EXOP format more closely by having the lone C bit in the middle. I can add something specifically mentioning the difference.

JKTKops:
What's the status after this one? https://discord.com/channels/828292123936948244/946806826756882552/1220820455963889685

I'm OK with this change if we just remove the alternative (i.e. current) layout entirely and maybe add some extra text instead.

EnderShadow:
What do you mean by "What's the status"? I don't think there's anything else that needs to be added to the PR.

Why do you think we should remove the current layout from this? What extra text would you suggest adding?

JKTKops:
re https://github.com/ETC-A/etca-spec/pull/72#issuecomment-2002074781
discussed briefly here https://discord.com/channels/828292123936948244/946806826756882552/1220820254343696574 I think it should be removed because I don't think it adds value but does add confusion.

`CCCC C CCCC` matches the actual spacing of the bits in the encoding, right? So I don't think the alternative is useful. If we're making a change like this for clarity, though, an example may help.

EnderShadow:
My reasoning for keeping the existing one is that it makes them a bit easier to parse when designing something that processes the already-extracted bits by being in 4-bit aligned chunks.

JKTKops:
I think it's confusing and unnecessary to have the same table repeated twice, even behind a spoiler. If one of them is easier to parse, then isn't that the one we should keep? In which case we could add an example encoding just to demonstrate that the way it turns into the table index is natural.

MegaIng:
I think adding a column with hex numbers equivalent might be helpful. That shouldn't take up too much space, is 4-bit aligned and is also useful when creating an assembler.

JKTKops:
I can get behind hex.

PR #73, opened 2024-03-16
Move instruction bits out of SS to ease decoding

EnderShadow:


JKTKops:
I'm not strictly opposed to this change, but I don't see how this simplifies even table-based decoders. Don't include those bits in your table index. Have the table identify this special format (just like it already has to identify other formats!) and then use these two bits to control the next layer.

EnderShadow:
If you think it doesn't improve it, then I'm fine leaving it as is since you're the one who's made a table based decoder.

JKTKops:
This change doesn't get rid of the second layer of table, if that's what you mean. It just changes which bits that table has to look at. This is still a special format that needs to be handled independently.

And actually, my second layer isn't a real table. It's just some equality checks to look for these 3 or 4 instructions.

EnderShadow:
In that case I think this is a helpful change. It allows the SS bits to be completely ignored for decoding the first byte if you have multiple first layer decoders based on the top bits.

JKTKops:
What would "multiple first layer decoders" even mean? I think you can ignore the SS bits anyway. You recognize this format by the pattern 00XX1111 in the first byte. This format doesn't make the SS bits relevant for recognizing _other_ formats, so you can just ignore them in the first layer.

EnderShadow:
Ah, I see now. you'd need to bring the SS bits into the second layer anyway for MM.

JKTKops:
The reason it's a "second layer" and not just part of a bigger first layer is because it's compartmentalized. The output of the first layer selects _which_ "second layer" to use. These instructions don't have MM bits, so "bring SS bits into the second layer for MM" doesn't really make sense.

My point is primarily that the second layer for these instructions doesn't really care where those two bits are. Having them in the SS bits does not pose any more or less complication than having them elsewhere, as far as I can tell. And it shouldn't matter for non-table based decoders either since these instructions really do have to be identified separately and don't interact with any SS-dependent components of a "model" backend.

PR #76, opened 2025-10-14
Initial Paging Proposals

JKTKops:
Many files have been changed or added.

Main changes:
* /extensions/mode-control-register.md: This document now describes addressing mode operation in significantly more detail and functions as a crash course in paging techniques for newcomers. Intended to be viewed with the table of contents tab open. Perhaps the name should be changed?
* /extensions/16-bit-paging/README.md: 16-bit paging extension.
* /extensions/32-bit-paging/README.md: 32-bit paging extension.
* /features/context-identifiers/README.md: CIDs for reducing virtual address space switch overhead.
* /features/page-global-enabled/README.md: Alternate, orthogonal method of reducing virtual address space switch overhead.

CIDs and global pages are complementary techniques. Every major ISA today has both.

The 16-bit paging extension has fairly minimal implementation requirements. A compliant implementation can get away with a 2x32-bit entry TLB; 4 new control registers, three of which are used to access the TLB entries, as well as one new bit in the `MODE` CR; one new exception type (though it is more complex than the most-similar `#MA`); and one new instruction, `INVLPG`, which _minimally_ did not need to be required but _practically_ is necessary to have.

A more maximal implementation of 16-bit paging is still missing a few real-life practical paging features, like Accessed and Dirty bits (though a method of emulating them is described in the document), an executable storage attribute, and global pages. It **does** support privileged access, read-only access, CIDs, a physical address space up to 4MiB in size, and hardware page table "walking."

The 32-bit paging extension is borderline identical to x86 paging without PAE. The key differences are a simplification of cache control bits (they are marked ignored in page directory entries as ARM/RISC-V have since proven that it is never necessary to have a write-through or uncachable page table), the inclusion of an executable storage attribute, the omission of Page Attribute Tables (the corresponding bit is reserved in case we want to add this in the future), and the shuffling of some bits to group ignored bits of page directories together so that the field is easier to use for software. A bit is reserved for using page directory entries to map "huge" 4MiB pages, but unlike CIDs and global pages, such huge pages are not fundamental to systems with paging and are left out of this initial proposal. It is also not clear how to modify the TLB entry format in the presence of huge pages, so I ensured that there was at least one bit reserved for making this work in the future.

It is specified that _all_ paging extensions will support operation with **either** hardware table walkers (and a correspondingly specified paging Radix Tree layout) or software TLB management (and a correspondingly specified TLB entry format). Extension documents are generally written in a way that preferences "hard paging" implementations. That is, the mechanics of the hard paging implementations are defined, and then the ways that "soft paging" differs are explained. It's _possible_ that this is an abstraction inversion, as first-time implementors will likely be focused on the behavior of soft paging.

The only remaining thing to pin down is where `TLBI` will live in the instruction encoding space. We may need to continue and finalize our discussion of a minor EXOP rework before that can happen. However, that's bikeshedding compared to everything else offered in this proposal, so let's discuss everything else for now.

These extensions are a long time coming. 😄

Supporting changes:
* /extensions/resources/vm-diagrams is now full of svg diagrams and sometimes the support files used to generate them. I was not very careful to save all of the support files but most of them are trivial to recreate. Sorry about that.
* moved /extensions/aoe-table into the new resources folder
* **Added /BRANCH-NOTES**, to be read by reviewers for context. This file will be deleted before merging.


JKTKops:
There are some things I'd like to discuss and I'll throw them on comments here as I think of them.

JKTKops:
Should soft paging implementations be allowed/required to support INVLPG and/or TLBI?

This creates a surprising amount of specification complexity. And, crucially, a hard paging implementation is allowed to receive INVLPG/TLBI and simply invalidate EVERYTHING, because its page table walkers can retrieve everything on-demand. A soft paging implementation can never be permitted to invalidate more than exactly what is specified, because of the risk of invalidating the TLB entry containing the page fault handler.

I think perhaps soft paging should be required to support INVLPG at least, so that someone could write a memory management "kernel library" that works for both soft and hard paging. I'm less set on TLBI which gets quite complicated I feel.

JKTKops:
I also keep confusing myself on what should happen when `VM_ROOT` is written to. Even for hard paging there is a tricky mess around whether or not CIDs are enabled: most of the time, invalidating _any_ entries while CIDs are enabled would defeat the point of CIDs, but when a CID is being recycled the previous entries for that CID must be discarded. X86-64 handles this with a downright mess; see page 5-48 of vol. 3A Intel® 64 and IA-32 Architectures Software Developer’s Manual. (In particular, the last bullet and its sub-bullets at the top of page 5-49.)

I think it might be simpler to say that (1) while CIDs are enabled, writes to `VM_ROOT` never invalidate anything. Use `TLBI` explicitly, which is required when CIDs are available anyways; and (2) when a write changes MODE[CID] from 1 to 0, all TLB entries except global ones are invalidated. (And of course while CIDs are disabled, writes to `VM_ROOT` must invalidate everything except global entries because `TLBI` might not be supported and there needs to be some way to invalidate everything.)

But even then, what about soft paging? The blanket invalidation when CIDs are disabled is obviously bogus (instant double fault on the next instruction fetch). Invalidating all non-global entries would work, except that it only works if the software is actually _using_ global pages -- it's not obligated to and in PG16 it _can't_. Because this is a problem when CIDs are **disabled** it is not easily solved by saying "use `TLBI`" as the instruction may not be available. Requiring the instruction would introduce a dependency on EXOP. Perhaps the solution is to say 'screw soft paging' and make them manually invalidate every TLB entry every time. I'm not opposed to this.

RISC-V dodges this problem by having the `TLBI` equivalent required by all implementations and by not specifying soft paging. Some RISC-V implementations support soft paging with an M ('machine')-mode exception in a non-standard way. I'd rather have it standardized but discouraged for ETCa.

JKTKops:
Another problem I've realized today: I've placed `INVLPG` in the non-canonical NOP region, but it **cannot** be silently ignored unlike the non-canonical NOP caching extensions. This means the placement is probably inappropriate and there isn't really anywhere else to put it without depending on EXOP. If we depend on EXOP anyway, `INVLPG` should just be fully replaced by `TLBI`, which also helps solve the above problem.

Thoughts?

JKTKops:
Another option is to reserve the `SS=00` cases from the EXOP jumps... and use them here (and also for `syscall`, `eret`, and `wait` while we're at it).

Then there would be no dependency on EXOP and `TLBI` could be the sole invalidation instruction. Also, it would still be 2 bytes, so there would be no dependency on VWI.

This gives me a stronger reason to invest in that possibility, so I will investigate table-based decoder complexity after updating the proposal with the address-related discussion we've recently had on Discord. https://discord.com/channels/828292123936948244/1435508980825849928

